PR #10, opened 2018-05-16
Task s1 handle login response

karimkefi:


public/js/loginValidation.js
```diff
@@ -17,30 +17,41 @@ let validateInputs = () => {
...
+    return response
+
 }
 
 let loginForm = document.getElementById('loginForm')
-loginForm.addEventListener('submit', (e) => {
-    let validInputs
+loginForm.addEventListener('submit', async (e) => {
     e.preventDefault()
+    let validInputs
```
nsantiagoblair: why are you assigning here and declaring on the following line?

public/js/loginValidation.js
```diff
@@ -17,30 +17,41 @@ let validateInputs = () => {
...
+            credentials: "same-origin",
             headers: {
                 'Accept': 'application/json',
-                'Content-Type': 'application/json'
+                'Content-Type': 'application/json',
             },
             method: "POST",
             body: JSON.stringify(data)
         })
         .then(function(data){return data.json()})
```
nsantiagoblair: ES6? 😿 
`.then(data => data.json())` beeeutiful!! 😸

public/js/loginValidation.js
```diff
@@ -17,30 +17,41 @@ let validateInputs = () => {
...
+        .then(function(data){return data})
+    return response
```
nsantiagoblair: could just return the fetch.. `return fetch()`
Only using the variable to return it

public/js/loginValidation.js
```diff
@@ -17,30 +17,41 @@ let validateInputs = () => {
     let cleanedPasswordInput = encodeURI(document.getElementById('password').value)
 
     if (isValidEmail('userEmail')) {
-        let emailPasswordValues = {'email' : cleanedEmailInput, 'password': cleanedPasswordInput}
+        let emailPasswordValues = {'userEmail' : cleanedEmailInput, 'password': cleanedPasswordInput}
         return emailPasswordValues
```
nsantiagoblair: Could just return the object. You're only using emailPasswordValues variable to return it...

public/js/loginValidation.js
```diff
@@ -17,30 +17,41 @@ let validateInputs = () => {
...
+loginForm.addEventListener('submit', async (e) => {
     e.preventDefault()
+    let validInputs
     validInputs = validateInputs()
-    sendLoginDetails(validInputs)
-})
+    let response = await sendLoginDetails('login', validInputs)
+
+    if(!response['success']) {
+        document.getElementById("error-message").innerText = response['msg']
```
nsantiagoblair: I don't think the user should be seeing the error message the fetch produces...

public/js/loginValidation.js
```diff
@@ -17,30 +17,41 @@ let validateInputs = () => {
...
+    } else if (response['success'] === true) {
+        window.location.href = "/admin";
+    }
```
nsantiagoblair: `let errorMsgBox = document.getElementById("error-message")`
`response['success'] === true ?`
 `window.location.href = "/admin" : errorMsgBox.innerText = response['msg']`
    🦅

templates/index.phtml
```diff
@@ -24,5 +24,6 @@
                 <span id="submitWarning" class="req"></span>
             </div>
         </form>
+        <div id="error-message"></div>
```
nsantiagoblair: if it pertains to the form i'd put this div inside the form

public/js/loginValidation.js
karimkefi: yup

public/js/loginValidation.js
karimkefi: ES5 ES5 ES5!!!

public/js/loginValidation.js
karimkefi: we need to await validate inputs, before awiting the fetch, before awaiting sendloginDetails, so cannot immediately return fetch()

public/js/loginValidation.js
karimkefi: fair point

public/js/loginValidation.js
karimkefi: sweet

templates/index.phtml
karimkefi: done

PR #23, opened 2018-11-26
Hiring partner tables

imogenhallett:
Please review sql in /db

### DIFF
--- a/db/academyPortal_2018-11-26.sql
+++ b/db/academyPortal_2018-11-26.sql
@@ -5,9 +5,9 @@
 # http://www.sequelpro.com/
 # https://github.com/sequelpro/sequelpro
 #
-# Host: 127.0.0.1 (MySQL 5.6.39)
+# Host: 127.0.0.1 (MySQL 5.6.41)
 # Database: academyPortal
-# Generation Time: 2018-05-18 08:30:31 +0000
+# Generation Time: 2018-11-26 15:33:13 +0000
 # ************************************************************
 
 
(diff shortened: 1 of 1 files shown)


db/academyPortal_2018-11-26.sql
```diff
@@ -20,6 +20,53 @@
...
+# ------------------------------------------------------------
+
+DROP TABLE IF EXISTS `hiringPartners`;
+
+CREATE TABLE `hiringPartners` (
+  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+  `companyName` varchar(255) NOT NULL DEFAULT '',
+  `size` int(11) unsigned NOT NULL,
+  `techStack` text NOT NULL,
+  `postcode` varchar(255) NOT NULL DEFAULT '',
```
hoptond: I don't think there's any postcode longer than 7 characters, 8 if you're including the space.

db/academyPortal_2018-11-26.sql
```diff
@@ -20,6 +20,53 @@
...
+  `phoneNo` varchar(11) DEFAULT NULL,
```
hoptond: Someone copying in a ++44 number as it appears will run out of characters before they get to the end of the number.

db/academyPortal_2018-11-26.sql
imogenhallett: Fixed in abf06b6

db/academyPortal_2018-11-26.sql
imogenhallett: Fixed in abf06b6

PR #29, opened 2018-11-27
Phb s3 t1 display applicants controller etc

mattap86:
created displayApplicantsController, Factory, added to route and dependencies

### DIFF
--- a/src/Classes/Factories/DisplayApplicantsControllerFactory.php
+++ b/src/Classes/Factories/DisplayApplicantsControllerFactory.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Portal\Factories;
+
+use Psr\Container\ContainerInterface;
+use Portal\Controllers\DisplayApplicantsController;
+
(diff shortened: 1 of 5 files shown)


src/Classes/Factories/DisplayApplicantsControllerFactory.php
```diff
@@ -0,0 +1,23 @@
...
+     * Instantiates DisplayApplicantsController with dependencies.
+     *
+     * @param ContainerInterface $container
+     *
+     * @return DisplayApplicantsController
+     */
+    public function __invoke(ContainerInterface $container) : DisplayApplicantsController
+    {
+        $renderer = $container->get('renderer');
+        $applicantModelFactory = $container->get('ApplicantModel');
```
maxmclarentaylor: change this to applicant model - otherwise it is a bit confusing - you have called on the factory to create the applicant model and so we are inserting the model not the factory

PR #43, opened 2018-11-29
S3 output data to event list

robertkieronreeves:
Pull request opened by Rob on behalf of Felix.

Display event data from DB into event list on account page (only events that haven't happened yet).


public/css/eventList.css
```diff
@@ -0,0 +1,80 @@
...
+    border: solid #fdedff 2px;
+}
+
+.top {
+    width: 90%;
+}
+
+.upper, .mid {
+    width: 100%;
+    background-color: #fbd8ff;
```
robertkieronreeves: Set a variable for background colour?

public/css/eventList.css
```diff
@@ -0,0 +1,80 @@
...
+
+.lower {
+    width: 100%;
+    background: #ffc0f1;
+}
+
+.arrowBox {
+    font-size: 2.5rem;
+    text-align: center;
+    background-color: #fbd8ff;
```
robertkieronreeves: As above. A variable for the colour could be set here.

public/css/eventList.css
```diff
@@ -0,0 +1,80 @@
...
+
+.addButton {
+    text-decoration: none;
+    border: solid #ffc0f1;
+    padding: 5px 30px;
+    border-radius: 5px;
+}
+
+.addButton:hover {
+    background-color: #fbd8ff;
```
robertkieronreeves: And again with regards to setting a colour variable.

public/css/eventList.css
```diff
@@ -0,0 +1,80 @@
...
+}
+
+.addButtonWrapper {
+    padding: 15px 0;
+    text-align: center;
+}
+
+.addButton {
+    text-decoration: none;
+    border: solid #ffc0f1;
```
robertkieronreeves: A colour variable could be set here too.

public/js/eventList.js
```diff
@@ -0,0 +1,36 @@
...
+        }
+        else {
+            arrow.innerHTML = "<span>&#x21E9;</span>"
+
+            var id = setInterval(function() {
+                if (lower.clientHeight > 0) {
+                    lower.style.height = (parseInt(lower.style.height) - 10) + 'px'
+                } else {
+                    clearInterval(id)
+                }
```
hoptond: These intervals cause very funky things to happen when the height of the element isn't divisible by 10 and/or there's lots of events on the screen at the same time.. Either fix this or use CSS transitions by applying a class with height: 0 when the info is hidden and height: auto when the info is visible.

src/Classes/Factories/EventListModelFactory.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Portal\Factories;
+use Portal\Models\EventListModel;
```
hoptond: insert newline between namespace and use

src/Classes/Models/EventListModel.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Portal\Models;
+
+
```
hoptond: too much whitespace here

src/Classes/ViewHelpers/EventListViewHelper.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Portal\ViewHelpers;
+
+class EventListViewHelper
+{
+    public static function outputEventList($events) {
+        foreach ($events as $event) {
+            $return =  '<div class= '. "event" . '>' .
```
hoptond: Not a good idea to name your variables after keywords.

PR #75, opened 2019-05-15
Rm pstory1 display info

mabuge90:


tests/Factories/GetApplicantControllerFactoryTest.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Tests\Factories;
+
+use PHPUnit\Framework\TestCase;
+
+use Portal\Factories\GetApplicantControllerFactory;
+use Portal\Models\ApplicationFormModel;
+use Psr\Container\ContainerInterface;
+
+
+class GetApplicantControllerFactoryTest extends TestCase
```
mporam: this test does not pass

public/js/applicantModal.js
```diff
@@ -1,80 +1,55 @@
...
-                        }
+                        console.log(data)
+                        validateField(data, 'name')
+                        validateField(data, 'email')
+                        validateField(data, 'phoneNumber')
+                        validateField(data, 'cohortDate')
+                        validateField(data, 'whyDev')
+                        validateField(data, 'codeExperience')
+                        validateField(data, 'hearAbout')
                         if (data.eligible !== 0) {
```
mporam: can you use the validateField function here with a custom message param

src/Classes/Controllers/GetApplicantController.php
```diff
@@ -30,8 +24,13 @@ public function __construct(ApplicantModel $applicantModel)
     public function __invoke(Request $request, Response $response, array $args)
     {
         $id = $args['id'];
-        $applicant = $this->applicantModel->getApplicantById($id);
-        return $response->withJson($applicant);
+        if(is_numeric($id) && !empty($id)){
+            $applicant = $this->applicantModel->getApplicantById($id);
+            return $response->withJson($applicant);
+        } else {
+            return $response->withJson('' , 418);
```
mporam: no

src/Classes/Entities/ApplicantEntity.php
```diff
@@ -59,21 +58,23 @@ public function __construct(
...
-                'codeExperience'=>$this->codeExperience,
-                'hearAboutId'=>$this->hearAboutId,
-                'eligible'=>$this->eligible,
-                'eighteenPlus'=>$this->eighteenPlus,
-                'finance'=>$this->finance,
-                'notes'=>$this->notes,
-                'cohortDate'=>$this->cohortDate,
-                'dateTimeAdded'=>$this->dateTimeAdded
-                ];
+        $formatedData = date("F, Y", strtotime($this->cohortDate) );
```
mporam: this could be in a separate method to make it reusable

mporam: infact `getCohortDate` does this for you, use that!

PR #79, opened 2019-05-15
Rm pstory2 task date filter

olyolyoly:


public/js/filterApplicantTable.js
```diff
@@ -0,0 +1,21 @@
+function filterByDate (dateFrom, dateTo){
+    document.querySelector('.btn.apply').addEventListener('click',function(){
+        document.querySelectorAll('#applicants .applicant').forEach(row => {
+
+            var dateCheck = row.querySelector('.dateApplied').dataset.applied;
```
alecorra: is a reason to use var rather than let?

public/js/filterApplicantTable.js
```diff
@@ -0,0 +1,21 @@
...
+
+            //console.log(dateCheck)
```
alecorra: clear the comment :)

src/Classes/Entities/ApplicantEntity.php
```diff
@@ -208,6 +210,22 @@ public function getNotes()
...
+     * Get's dateOfApplication.
+     *
+     * @return string, returns the dateOfApplication field.
+     */
+    public function getDateOfApplication()
+    {
+        return $this->dateTimeAdded;
+    }
+
+    public function getDateOfApplicationUsingMMDDYYYY()
```
alecorra: could this function be named simpler?

src/Classes/Entities/ApplicantEntity.php
```diff
@@ -217,4 +235,6 @@ public function getCohortDate()
     {
         return $this->cohortDate;
     }
+
```
alecorra: clear the whitespace

PR #81, opened 2019-05-15
Rm pstory3 task static search

Liliya-V:


templates/displayApplicants.phtml
```diff
@@ -14,6 +14,12 @@
 </head>
 <body>
     <div class="container">
+        <div class="row">
+            <!-- Search form -->
```
lucyholubova: maybe we dont need this as the search bar is obvious

PR #86, opened 2019-05-15
added hotfix to improve phone regex

JKapella:


Knux2:
It does the thing. Awesome.

PR #89, opened 2019-05-15
Rm pstory2 task cohort filter

alecorra:


public/js/cohortFilter.js
```diff
@@ -0,0 +1,33 @@
...
+            let row = date.parentNode
+            mismatches.push(row)
+        }
+    })
+    mismatches.forEach(function(mismatch) {
+        mismatch.classList.add('hidden')
+    })
+}
+
+function clearFilter() {
```
Liliya-V: Dock block?

public/js/cohortFilter.js
```diff
@@ -0,0 +1,33 @@
+let tableRows = document.querySelectorAll('.table-row')
+let applicantsDates = document.querySelectorAll('.applicants-cohort-date')
+
+function applyCohortFilter(cohortDate) {
```
Liliya-V: Dockblock?

public/js/cohortFilter.js
alecorra: done

public/js/cohortFilter.js
alecorra: done

PR #102, opened 2019-05-16
Task3 5 hiring partner contacts model

katestarks:


src/Classes/Models/HiringPartnerContactsModel.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        $this->db = $db;
+    }
+
+    /**
+     * Queries the database to get an array of hiring partner company ids
+     *
+     * @return array
+     */
+    public function getHiringPartnerContactById() :array {
```
aldofiore28: Is this method supposed to get something by id? There is not input and in the statement, and you take everything from the database, not a specific one.

src/Classes/Models/HiringPartnerContactsModel.php
```diff
@@ -0,0 +1,73 @@
...
+
+    private $db;
+
+    public function __construct(\PDO $db)
+    {
+        $this->db = $db;
+    }
+
+    /**
+     * Queries the database to get an array of hiring partner company ids
```
aldofiore28: Do you just get the IDs?

src/Classes/Models/HiringPartnerContactsModel.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+namespace Portal\Models;
+
+use Portal\Entity\HiringPartnerContactEntity;
+
+class HiringPartnerContactsModel
+{
+
```
aldofiore28: remove linebreak please

PR #105, opened 2019-05-16
Task3 6 hiring partner contact model factory

rrrichard:


src/Classes/Factories/HiringPartnerContactsModelFactory.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Portal\Factories;
+
+use Psr\Container\ContainerInterface;
+use Portal\Models\HiringPartnerContactsModel;
+
+class HiringPartnerContactsModelFactory
+{
+    /** Calls addHiringPartnerContact method
```
aldofiore28: Move this to the next line

src/Classes/Factories/HiringPartnerContactsModelFactory.php
```diff
@@ -0,0 +1,21 @@
...
+     *
+     * @param $container
+     *
+     * @return HiringPartnerContactsModel returns JSON
```
aldofiore28: It does not return JSON, it returns a new instantiation of the model

PR #124, opened 2019-09-30
added async and await to click event listener

MotNosniktaw:


public/js/hiringPartner.js
```diff
@@ -1,11 +1,11 @@
 
-document.getElementById('submitHiringPartner').addEventListener('click', e => {
+document.getElementById('submitHiringPartner').addEventListener('click', async e => {
```
mallowmew: Put `e` in brackets for consistency with line 89

PR #126, opened 2019-09-30
Axolotls Story 1 - Fix Applicant Info Page

mallowmew:


mallowmew:
Wait for S1T2

lukestur:
NB - the displayApplicants page does not currently require you to be logged in. Is this an oversight or by design??

PR #127, opened 2019-09-30
S3T4 - Update Database structure

mallowmew:


db/academyPortal_2019-09-30.sql
```diff
@@ -115,6 +115,51 @@ VALUES
...
+  `category` int(11) unsigned NOT NULL,
+  `date` date NOT NULL,
+  `start_time` time NOT NULL,
+  `end_time` time NOT NULL,
+  `notes` text,
+  PRIMARY KEY (`id`),
+  KEY `category` (`category`),
+  CONSTRAINT `events_ibfk_1` FOREIGN KEY (`category`) REFERENCES `event_categories` (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
```
lizsharp: Need to add a location field.

lizsharp: 'Location' is also missing from the getEvents method in the eventModel. This needs to be added.

mallowmew: Have done this in separate fix branch

PR #129, opened 2019-09-30
S3T3 Create Event Model

mallowmew:


MotNosniktaw:
oops! add eventmodel test class with contructor test

PR #131, opened 2019-10-01
S3T6 AddEvent Route

mallowmew:


src/Classes/Controllers/AddEventController.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Portal\Controllers;
+
+use Slim\Http\Request;
+use Slim\Http\Response;
+use Slim\Views\PhpRenderer;
+use Portal\Models\EventModel;
+use Portal\Entities\EventEntity;
+
+class AddApplicantController
```
MotNosniktaw: should be AddEventController

src/Classes/Controllers/AddEventController.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    private $eventModel;
+
+    /**
+     * AddApplicantController Constructor
```
MotNosniktaw: should reflect adjusted class name

src/Classes/Controllers/AddEventController.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param EventModel $eventModel
+     */
+    public function __construct(EventModel  $eventModel)
+    {
+        $this->eventModel = $eventModel;
+    }
+
+    public function __invoke(Request $request, Response $response, array $args)
```
MotNosniktaw: return type is Response object

PR #139, opened 2019-10-01
S3t11 backend validation for new tasks

AlisonYoon:


src/Classes/Entities/EventEntity.php
```diff
@@ -4,6 +4,7 @@
 
 class EventEntity
 {
+    protected $id;
```
mallowmew: Why did you add id? The id is generated by the database, not supplied by the user.

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
+    }
+
+    /**
+     * Will sanitise all the fields for an applicant.
+     */
+    private function sanitiseData()
+    {
+        $this->id = (int) $this->id;
+        $this->name = $this->sanitiseString($this->name);
+        $this->name = self::validateLength($this->name, 20);
```
mallowmew: Any reason to limit length to 20 rather than 255?

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
+        $this->category = (int)$this->category;
+        $this->location = $this->sanitiseString($this->location);
+        $this->location = self::validateLength($this->location, 255);
+        $this->date = $this->sanitiseString($this->date);
```
mallowmew: The dates and times should use some regex to make sure that they look like dates and times in the format that mySQL will be expecting e.g. `'2019-10-02'`, `'16:55'`

src/Classes/Entities/EventEntity.php
AlisonYoon: Yeah I was confused if I should add this or not, seeing that they did in the ApplicantEntity, I did the same. I will remove it.

src/Classes/Entities/EventEntity.php
AlisonYoon: valid point

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
+        $this->sanitiseData();
+    }
+
+    /**
+     * Will sanitise all the fields for an applicant.
+     */
+    private function sanitiseData()
+    {
+        $this->id = (int) $this->id;
+        $this->name = $this->sanitiseString($this->name);
```
mallowmew: All of these string fields should probably have length checks

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
+        $this->name = self::validateLength($this->name, 20);
+        $this->category = (int)$this->category;
+        $this->location = $this->sanitiseString($this->location);
+        $this->location = self::validateLength($this->location, 255);
+        $this->date = $this->sanitiseString($this->date);
+        $this->startTime = $this->sanitiseString($this->startTime);
+        $this->endTime = $this->sanitiseString($this->endTime);
+        $this->notes = $this->sanitiseString($this->notes);
```
mallowmew: This should probably have a length check too.

AlisonYoon: exist and length check maybe?

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
         $this->date = ($this->date ?? $date);
         $this->startTime = ($this->startTime ?? $startTime);
         $this->endTime = ($this->endTime ?? $endTime);
         $this->notes = ($this->notes ?? $notes);
+
+        $this->sanitiseData();
+    }
+
+    /**
+     * Will sanitise all the fields for an applicant.
```
lizsharp: for an event not an applicant

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
+        $this->location = $this->sanitiseString($this->location);
+        $this->location = self::validateLength($this->location, 255);
+        $this->date = $this->sanitiseString($this->date);
+        $this->startTime = $this->sanitiseString($this->startTime);
+        $this->endTime = $this->sanitiseString($this->endTime);
+        $this->notes = $this->sanitiseString($this->notes);
+    }
+
+    /**(
+     * Sanitise as a string in the applicant table as data.
```
lizsharp: event/applicant

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
+     *
+     * @param string $eventData
+     *
+     * @return string, which will return the applicant data.
```
lizsharp: event/applicant

src/Classes/Entities/EventEntity.php
```diff
@@ -13,20 +14,170 @@ class EventEntity
...
+            return $eventData;
+        } else {
+            throw new \Exception('An input string does not exist or is too long');
+        }
+    }
+
+    /**
+     * Validate that a string is not empty and is within length allowed, throws an error if not
+     *
+     * @param string $hiringPartnerData
```
lizsharp: eventdata not hiringpartnerdata

src/Classes/Entities/EventEntity.php
mallowmew: Notes doesn't need to exist. Other fields do.

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,173 @@ public function __construct(
...
+     */
+    public function sanitiseDate($date)
+    {
+        if (!preg_match('/([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/', $date)) {
+            return false;
+        } else {
+            return $this->sanitiseString($this->date);
+        }
+    }
+
```
mallowmew: doc block?

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,173 @@ public function __construct(
...
+        $this->sanitiseData();
+    }
+
+    /**
+     * Will sanitise all the fields for an event.
+     */
+    private function sanitiseData()
+    {
+        $this->name = $this->sanitiseString($this->name);
+        $this->name = self::validateLength($this->name, 255);
```
mallowmew: Required fields should probably validate that they exist and are within length

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,173 @@ public function __construct(
...
+        $this->category = (int)$this->category;
```
mallowmew: Would it be possible to check if the category exists in the database?

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,173 @@ public function __construct(
...
+    }
+
+    /**
+     * Sanitise as a date in the event table as data.
+     *
+     * @param $date
+     *
+     * @return bool|string
+     */
+    public function sanitiseDate($date)
```
mallowmew: This and the time functions I would say are validating rather than sanitising - sanitisation normally means 'making the same data cleaner' while these are checking that the data matches an expected pattern.

src/Classes/Entities/EventEntity.php
AlisonYoon: ok

src/Classes/Entities/EventEntity.php
AlisonYoon: ok

src/Classes/Entities/EventEntity.php
AlisonYoon: ok

src/Classes/Entities/EventEntity.php
AlisonYoon: yes but not in a clean way

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,179 @@ public function __construct(
...
+     * Sanitise as a date in the event table as data.
+     *
+     * @param $date
+     *
+     * @return bool|string
+     */
+    public function validateDate($date)
+    {
+        if (!preg_match('/([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/', $date)) {
+            return false;
```
mallowmew: This will assign a value of false to the date property of the entity - what then happens to that value of false on the way to the database?

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,179 @@ public function __construct(
...
+    }
+
+    /**
+     * Sanitise as a date in the event table as data.
+     *
+     * @param $date
+     *
+     * @return bool|string
+     */
+    public function validateDate($date)
```
FelixLetheren: Could $date be typehinted?

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,179 @@ public function __construct(
...
+     *
+     * @param string $eventData
+     *
+     * @return string, which will return the event data.
+     */
+    public function sanitiseString($eventData)
+    {
+        return filter_var($eventData, FILTER_SANITIZE_STRING);
+    }
+
```
MotNosniktaw: why use a named function that only contains a builtin function?

AlisonYoon: this named function was used in other entities so I just did the same how it was done before.

src/Classes/Entities/EventEntity.php
AlisonYoon: changed to throw Exception

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,179 @@ public function __construct(
...
+    {
+        if (!preg_match('/((1[0-2]|0?[1-9]):([0-5][0-9]))/', $time)
+        ) {
+            throw new \Exception('Please enter correct time');
+        } else {
+            if ($time == $this->startTime) {
+                return $this->sanitiseString($this->startTime);
+            } elseif ($time == $this->endTime) {
+                return $this->sanitiseString($this->endTime);
+            }
```
FelixLetheren: This doesn't cover every eventuality as there isn't any return if none of these requirments are met.

src/Classes/Entities/EventEntity.php
```diff
@@ -84,7 +84,7 @@ public function validateTime(string $time)
 
     public function validateStartEndTime($startTime, $endTime)
     {
-        if($startTime >= $endTime) {
+        if ($startTime >= $endTime) {
             throw new \Exception('End time should be later than Start time');
```
FelixLetheren: Should the S of start be capatalised?

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,185 @@ public function __construct(
...
+        $this->endTime = $this->validateTime($this->endTime);
+        $this->validateStartEndTime($this->startTIme, $this->endTime);
+        $this->notes = $this->sanitiseString($this->notes);
+        $this->notes = self::validateLength($this->notes, 255);
+    }
+
+    /**
+     * Sanitise as a date in the event table as data.
+     *
+     * @param $date
```
FelixLetheren: doc bock not up to date with type hinting.

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,185 @@ public function __construct(
...
+            throw new \Exception('Please enter correct date');
+        } else {
+            return $date;
+        }
+    }
+
+    /**
+     * Sanitise as a time in the event table as data.
+     *
+     * @param $time
```
FelixLetheren: doc block not up to date with type hinting

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,185 @@ public function __construct(
...
+    public function validateTime(string $time)
+    {
+        if (!preg_match('/((1[0-2]|0?[1-9]):([0-5][0-9]))/', $time)
+        ) {
+            throw new \Exception('Please enter correct time');
+        } else {
+            return $time;
+        }
+    }
+
```
FelixLetheren: missing docblock

src/Classes/Entities/EventEntity.php
```diff
@@ -28,5 +28,185 @@ public function __construct(
...
+    public function validateStartEndTime($startTime, $endTime)
+    {
+        if ($startTime >= $endTime) {
+            throw new \Exception('End time should be later than start time');
+        } else {
+            return true;
+        }
+    }
+
+    /**(
```
FelixLetheren: what's this little ( doing over here? Cheeky little fella.

AlisonYoon: respect

FelixLetheren:
**wowee**

PR #140, opened 2019-10-01
S3T9 JS to Add new Events

mallowmew:


public/js/addEvent.js
```diff
@@ -0,0 +1,43 @@
...
+            eventForm.elements['event-location'].value = '',
+            eventForm.elements['event-date'].value = '',
+            eventForm.elements['event-start-time'].value = '',
+            eventForm.elements['event-end-time'].value = '',
+            eventForm.elements['event-notes'].value = ''
+            message.innerText = responseJson.message
+        } else {
+            message.innerText = responseJson.message
+        }
+    })
```
MotNosniktaw: placeholder function name?

public/js/addEvent.js
```diff
@@ -0,0 +1,43 @@
+const eventList = document.querySelector('#events')
+const eventForm = document.querySelector('form')
+const message = document.querySelector('#messages')
```
MotNosniktaw: swap queryselector for getElementByID?

public/js/addEvent.js
mallowmew: Let's do it in a JS cleanup task

PR #141, opened 2019-10-01
S3t10 front end validation

FelixLetheren:
Added validation to new events form before is converted to an object.

### DIFF
--- a/public/js/EventsPage.js
+++ b/public/js/EventsPage.js
@@ -32,7 +32,7 @@ function validateForm() {
             message += 'Please select an event category!<br>'
             success = false
         }
-        //Checks the date is in the format of a date
+        //Checks the date is in the format of a date 'YYYY-MM-DD'
         if (element.name === 'event-date') {
             let date =  element.value.trim()
             let pattern = /([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/
@@ -42,7 +42,7 @@ function validateForm() {
                 success = false
             }
         }
-        //Checks the start time is in the format of a time
+        //Checks the start time is in the format of a time 'HH:MM'
         if (element.name === 'event-start-time') {
             let date =  element.value.trim()
             let pattern = /([01]?[0-9]|2[0-3]):[0-5][0-9](:[0-5][0-9])?/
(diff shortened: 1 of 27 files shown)


public/js/EventsPage.js
```diff
@@ -0,0 +1,78 @@
...
+    let success = true
+    let message = ''
+    let inputs = document.querySelectorAll('.create-events')
+    inputs.forEach(function (element) {
+        //Checks fields with attribute data-required=true has data
+        let required = element.getAttribute('data-required')
+        if (required && element.value.length < 1) {
+            message += element.previousElementSibling.innerHTML + ' is a required field! <br>'
+            success = false
+        }
```
mallowmew: In the interests of DRY and code-not-falling-over, I would have a separate validation function for each field. This way if there's a problem with the validation across mutiple fields all of those errors can be shown at once / not blocked waiting for previous fields to validate.

public/js/EventsPage.js
```diff
@@ -0,0 +1,78 @@
...
+        if (required && element.value.length > maxLength) {
+            message += element.name + ' is too long, must be less than ' + maxLength + ' characters! <br>'
+            success = false
+        }
+        //Checks an event is selected
+        if (element.name === 'event-category' && element.value === '0') {
+            message += 'Please select an event category!<br>'
+            success = false
+        }
+        //Checks the date is in the format of a date
```
mallowmew: Give an example in the comment of the date format expected

public/js/EventsPage.js
```diff
@@ -0,0 +1,78 @@
...
+        if (element.name === 'event-date') {
+            let date =  element.value.trim()
+            let pattern = /([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/
+            let regEx = new RegExp(pattern)
+            if (!regEx.test(date)) {
+                message += 'Invalid date!<br>'
+                success = false
+            }
+        }
+        //Checks the start time is in the format of a time
```
mallowmew: Give an example in the comment of the time format expected

public/js/EventsPage.js
```diff
@@ -0,0 +1,78 @@
...
+                message += 'Invalid end time!<br>'
+                success = false
+            }
+        }
+    })
+    //Adds all error messages to the messages div.
+    document.getElementById('messages').innerHTML = message
+    return success
+}
+//Adds data from form into an object with the field name as key and the form value as value.
```
mallowmew: Consider a doc block

public/js/EventsPage.js
FelixLetheren: being done on seperate branch

PR #143, opened 2019-10-01
S3t8 get+display events

RachmannJoubert:
Functions for retireving the events information from the database and displaying it on the vents page.

### DIFF
--- a/public/js/getEvents.js
+++ b/public/js/getEvents.js
@@ -1,4 +1,4 @@
-async function getEvents () {
+async function getEvents() {
     await fetch('/api/getEvents', {
         credentials: "same-origin",
         headers: {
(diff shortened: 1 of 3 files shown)


public/js/getEvents.js
```diff
@@ -0,0 +1,58 @@
...
+        .then(eventInfo => eventInfo.json())
+        .then(eventInfo => displayEventsHandler(eventInfo.data))
+}
+
+/**
+ * Runs a foreach through each event object and outputs HTML elements with event details
+ *
+ * @param events is an array of objects which contains information about events
+ *
+ * @return a divs of the event name with a button that reveals each event's additional info on each line
```
FelixLetheren: should this be 'div'? Doesn't make sense to me.

public/js/getEvents.js
```diff
@@ -0,0 +1,58 @@
...
+        },
+    })
+        .then(eventInfo => eventInfo.json())
+        .then(eventInfo => displayEventsHandler(eventInfo.data))
+}
+
+/**
+ * Runs a foreach through each event object and outputs HTML elements with event details
+ *
+ * @param events is an array of objects which contains information about events
```
FelixLetheren: should be typehinted to array?

mallowmew: can't type hint in js

public/js/getEvents.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Gets event information from the API and passes into the displayHandler function
+ *
+ * @return event data
+ */
+async function getEvents() {
```
mallowmew: Does not need to be async/await

public/js/getEvents.js
mallowmew: It doesn't return anything, it spits out html

templates/EventsPage.phtml
```diff
@@ -41,7 +42,7 @@
         <label for='event-notes'>Notes</label>
         <input id='event-notes' name='event-notes' class="create-events form-control" type="text" placeholder="Notes"
                data-max='255'>
-        <input id="submit-event" type="submit" class="btn btn-success"/>
+        <input id="create-events" type="submit" class="btn btn-success"/>
```
mallowmew: Why did this change?

templates/EventsPage.phtml
```diff
@@ -41,7 +42,7 @@
         <label for='event-notes'>Notes</label>
         <input id='event-notes' name='event-notes' class="create-events form-control" type="text" placeholder="Notes"
                data-max='255'>
-        <input id="submit-event" type="submit" class="btn btn-success"/>
+        <input id="submit-events" type="submit" class="btn btn-success"/>
```
mallowmew: You submit a single event at a time

PR #153, opened 2019-10-02
Changed formatting of time to display without seconds.

FelixLetheren:


public/js/eventsPage.js
```diff
@@ -30,6 +30,8 @@ function getEvents() {
 function displayEventsHandler(events) {
     let eventInformation = ''
     events.forEach(event => {
+        let startTimeWithoutSeconds = event.start_time.slice(0, - 3)
+        let endTimeWithoutSeconds = event.end_time.slice(0, - 3)
```
MotNosniktaw: consider keeping YYYY-MM-DD rather than cutting SS. I am happy, your description

mallowmew: wat r u on about lol

MotNosniktaw: hello. i am potato

PR #158, opened 2019-10-02
Fix events page+events controller message

RachmannJoubert:


public/js/eventsPage.js
```diff
@@ -29,7 +29,14 @@ function getEvents() {
  */
 function displayEventsHandler(events) {
     let eventInformation = ''
+    let noEvents = document.querySelector('.no-events')
```
FelixLetheren: could go straight into the evets div rather than having another span. Just a bit cleaner.

src/Classes/Controllers/GetEventsController.php
```diff
@@ -51,6 +51,13 @@ public function __invoke(Request $request, Response $response, array $args)
                 'data' => $events
             ];
             $statusCode = 200;
+        } else {
+            $data = [
+                'success' => true,
+                'message' => 'There is no event',
```
mallowmew: 'There are no events.'

src/Classes/Controllers/GetEventsController.php
```diff
@@ -51,6 +51,13 @@ public function __invoke(Request $request, Response $response, array $args)
                 'data' => $events
             ];
             $statusCode = 200;
+        } else {
```
FelixLetheren: isn't this doing the same as line 34-39?

src/Classes/Controllers/GetEventsController.php
FelixLetheren: *There are no events

src/Classes/Controllers/GetEventsController.php
```diff
@@ -51,6 +51,13 @@ public function __invoke(Request $request, Response $response, array $args)
...
+            $data = [
+                'success' => true,
+                'message' => 'There is no event',
+                'data' => []
+            ];
+            $statusCode = 200;
```
FelixLetheren: Could be code 204
"204 No Content
    The server successfully processed the request and is not returning any content."

FelixLetheren:
SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data

PR #160, opened 2019-10-02
created ValidationEntity abstract class that is a parent class that c…

AlisonYoon:
…ontains all the common methods

### DIFF
--- a/src/Classes/Entities/ValidationEntity.php
+++ b/src/Classes/Entities/ValidationEntity.php
@@ -48,7 +48,7 @@ abstract class ValidationEntity
      *
      * @return string, which will return the validateData.
      */
-    public function sanitiseString(string $validateData)
+    public static function sanitiseString(string $validateData)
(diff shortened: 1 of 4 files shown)


src/Classes/Entities/ValidationEntity.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
```
Tibereon: Extra lines

src/Classes/Entities/ValidationEntity.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Sanitise as a string in the database table as data.
+     *
+     * @param string $validateData
+     *
+     * @return string, which will return the validateData.
+     */
+    public function sanitiseString(string $validateData)
```
mallowmew: Could this also be a static function?

PR #164, opened 2019-10-03
S5t3 add contact api route

RachmannJoubert:


src/Classes/Controllers/AddContactController.php
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * AddContactController constructor.
+     * @param $hiringPartnerModel
+     */
+    public function __construct($hiringPartnerModel)
+    {
+        $this->hiringPartnerModel = $hiringPartnerModel;
+    }
+
+
```
mallowmew: You need an invoke too - the method on the model you'll need to call is `addNewContact`. 
Look at `CreateHiringPartnerController` or `AddEventController` for examples.

src/Classes/Controllers/AddContactController.php
```diff
@@ -0,0 +1,59 @@
...
+            if (!empty($contact) && $contact instanceof HiringPartnerEntity) {
+                $result = $this->hiringPartnerModel->addNewContact($contact);
+            }
+        } catch (\Exception $exception) {
+            $data['message'] = $exception->getMessage();
+        }
+        if (isset($result) && $result) {
+            $data = [
+                'success' => true,
+                'message' => 'Hiring Partner Added to the db',
```
AlisonYoon: Should probably say something like 'Hiring Parter Contact added to the db' ?

src/Classes/Controllers/AddContactController.php
```diff
@@ -0,0 +1,59 @@
...
+            $contact = $this->hiringPartnerModel->addNewContact(
+                $newContact['name'],
+                $newContact['email'],
+                $newContact['job_title'],
+                $newContact['phone'],
+                $newContact['hiring_partner_company_id'],
+                $newContact['is_primary_contact']
+            );
+            if (!empty($contact) && $contact instanceof HiringPartnerEntity) {
+                $result = $this->hiringPartnerModel->addNewContact($contact);
```
AlisonYoon: I don't see addNewContact() in the hiringPartnerModel

RachmannJoubert: The method will be created in a seperate task.

AlisonYoon: nvm that's in another task

src/Classes/Controllers/AddContactController.php
```diff
@@ -0,0 +1,59 @@
...
+        $statusCode = 400;
+
+        try {
+            $contact = $this->hiringPartnerModel->addNewContact(
+                $newContact['name'],
+                $newContact['email'],
+                $newContact['job_title'],
+                $newContact['phone'],
+                $newContact['hiring_partner_company_id'],
+                $newContact['is_primary_contact']
```
mallowmew: Can we have these fields named 

```
contactName
contactCompanyId
contactEmail
contactJobTitle
contactPhone
contactIsPrimary
```

PR #166, opened 2019-10-03
hiring parter company dropdown added

AlisonYoon:


src/Classes/Models/HiringPartnerModel.php
```diff
@@ -56,6 +56,17 @@ public function getCompanySize() :array
...
+     *
+     * @return array
+     */
+    public function getCompanyName() :array
+    {
+        $query = $this->db->prepare("SELECT `id`,`name` FROM `hiring_partner_companies`");
+        $query->execute();
+        return $query->fetchAll();
+    }
+
```
mallowmew: Can you give this to Liz so it can be part of her task and remove it from this branch?

AlisonYoon: sure

PR #167, opened 2019-10-03
S5T8 & S5T5 Add Contact Frontend & Validation

mallowmew:


mallowmew:
Thsi should all be done but breaks because the model doesn't have a method yet :)

public/js/validations.js
```diff
@@ -14,4 +14,41 @@ function isTime(time) {
...
+    let regEx = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+    if (regEx.test(email)) {
+        return true
+    } else {
+        return false
+    }
+}
+
+function isUrl(url) {
+    let regEx = /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$/gm
```
AlisonYoon: should <http(s)> be optional or mandatory? On company's name card or something, many places omit http(s) in front of the website url.

public/js/validations.js
```diff
@@ -14,4 +14,41 @@ function isTime(time) {
...
+    if (regEx.test(phone)) {
+        return true
+    } else {
+        return false
+    }
+}
+
+function isEmail(email) {
+    // email regex from http://emailregex.com - "Email Address Regular Expression That 99.99% Works."
+    let regEx = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
AlisonYoon: you can have periods in id, but after @, there should be limit in number of periods isn't it? 
(sdfsdfsdfadf@sdfsdf.dsf.sdfs.sdf.asdf.asdf.asdf.asdf) -> this weird address passes the email validation

public/js/validations.js
mallowmew: It's optional but we might consider trimming or enforcing it in the backend

public/js/validations.js
mallowmew: sdfsdfsdfadf@sdfsdf.dsf.sdfs.sdf.asdf.asdf.asdf.asdf could actually be a standard-compliant email address ¯\_(ツ)_/¯

public/js/validations.js
```diff
@@ -14,4 +14,41 @@ function isTime(time) {
     } else {
         return false
     }
+}
+
+function isPhoneNumber(phone) {
+    let regEx = /^(([+][(]?[0-9]{1,3}[)]?)|([(]?[0-9]{4}[)]?))\s*[)]?[-\s\.]?[(]?[0-9]{1,3}[)]?([-\s\.]?[0-9]{3})([-\s\.]?[0-9]{3,4})$/gm
```
AlisonYoon: after the country code, if the first digit is 0 then you should omit it, so for example it goes:
+44(0)7100123345
Does this kind of filtering for display be done in this branch do you think?

public/js/validations.js
AlisonYoon: for now, without http(s), it doesn't pass the regex validation.
for instance something like (www.mayden.com) doesn't pass this.

public/css/genericStyles.css
```diff
@@ -43,7 +43,7 @@ input[type="submit"] {
     border-top: 1px solid #CCCCCC
 }
 
-.moreInfo {
+.more-info {
```
AlisonYoon: I think this was done in Matt's branch but margin-top and margin-bottom doesn't have to be in two separate lines. 
margin: 10px 0;
should do the job.

PR #172, opened 2019-10-03
Axolotls Story 5 Review - Add Contacts to Hiring Partners

mallowmew:


mallowmew:
The page for adding new applicants doesn't work, and we didn't touch the code for that feature this sprint. This problem might not have been noticed previously because the absolute paths in the application (which we fixed in one of our stories this sprint) broke the page in a different way. 

THe problem seems to be that there's a type-cast missing when the ApplicantEntity is constructed, which throws back a Slim Exception.

PR #176, opened 2019-11-18
Add sort buttons

jasonrees1:


templates/displayApplicants.phtml
```diff
@@ -20,8 +21,8 @@
                 <tr>
                     <th class="col-xs-2">Name</th>
                     <th class="col-xs-4">Email</th>
-                    <th class="col-xs-2">Date of application</th>
-                    <th class="col-xs-2">Cohort date</th>
+                    <th class="col-xs-2">Date of application <i style="color:#8d0377;" class="fas fa-arrow-up"></i> <i class="fas fa-arrow-down"></i></th>
```
georginaw: The formatting of the arrows is not super clear - purple is too close to black I think, could they also be made bigger and be positioned so that they're consistent across both?

PR #182, opened 2019-11-19
G1 t3 link hp to event controller

bentleybiscuit:


src/Classes/Controllers/LinkHiringPartnerToEventController.php
```diff
@@ -0,0 +1,42 @@
...
+     * @param Response $response HTTP response
+     * @param array $args
+     *
+     * @return Response returns Json success/failure message
+     */
+    public function __invoke(Request $request, Response $response, array $args) :Response
+    {
+        $data = $request->getParsedBody();
+        $hiringPartner = $data['companyId'];
+        $event = $data['eventId'];
```
jdselby24: Change these keys to match the DB field names

src/Classes/Controllers/LinkHiringPartnerToEventController.php
```diff
@@ -0,0 +1,42 @@
...
+        $attendees = $data['attendees'];
+        $result = $this->eventModel->linkHPToEvent($hiringPartner, $event, $attendees);
+        if ($result) {
+            return $response->withJson(['success' => 'True'], 200);
+        } else {
+            return $response->withJson(['success' => 'False'], 200);
```
jdselby24: Should respond with a status code other than 200, because if its failed, it's not 200 OK
```suggestion
            return $response->withJson(['success' => 'False'], 500);
```

src/Classes/Entities/EventEntity.php
```diff
@@ -14,6 +15,7 @@ class EventEntity extends ValidationEntity
     protected $eventCategories;
 
     public function __construct(
+        string $eventId = null,
```
jdselby24: should match db

src/Classes/Entities/EventEntity.php
```diff
@@ -23,6 +25,7 @@ public function __construct(
         string $notes = null,
         array $eventCategories
     ) {
+        $this->eventId = ($this->eventId ?? $eventId);
```
jdselby24: should match db

src/Classes/Entities/EventEntity.php
```diff
@@ -40,6 +43,7 @@ public function __construct(
      */
     private function sanitiseData()
     {
+        $this->eventId = (int)$this->eventId;
```
jdselby24: should match db

src/Classes/Entities/EventEntity.php
```diff
@@ -120,6 +124,17 @@ public static function validateCategoryExists(int $category, array $categoryList
...
+     * Get event id
+     *
+     * @return string id
+     */
+    public function getEventId(): int
+    {
+        return $this->eventId;
+    }
+
+
```
jdselby24: should match db

src/Classes/Entities/EventEntity.php
```diff
@@ -4,6 +4,7 @@
 
 class EventEntity extends ValidationEntity
 {
+    protected $eventId;
```
jdselby24: should match db

src/Classes/Controllers/LinkHiringPartnerToEventController.php
jdselby24: Consider sending a message back as well

PR #184, opened 2019-11-19
C2t2 validate event search

georginaw:
Created front end (JS and HTML tag) and back end validation (in the get events controller invoke function) to check if the user has put in a value in the input and that it's not longer than 255 characters


public/js/eventsPage.js
```diff
@@ -8,8 +8,13 @@ const message = document.querySelector('#messages')
  *
  * @return event data
  */
-function getEvents() {
-    fetch('./api/getEvents', {
+function getEvents(search = false) {
+    let url = './api/getEvents'
+    if (search !== false) {
```
anton25360: Why did you use !== false rather than == true?.

public/js/eventsPage.js
```diff
@@ -180,4 +185,15 @@ function validateForm() {
...
     return success
-}
\ No newline at end of file
+}
+
+
+document.querySelector('#submit-search-event').addEventListener('click', function(e) {
+    const searchInput = document.querySelector('#academy-events-search').value
+    e.preventDefault()
+    if ((searchInput.length) && searchInput.length < 255) {
```
anton25360: Check '255'. It should probably be 256, just like in this line:
`if (strlen($eventSearchInput) < 256) {`

public/js/eventsPage.js
georginaw: This is a strict comparison and makes more sense than using true (used when the input it blank)

public/js/eventsPage.js
georginaw: Updated to 256

PR #188, opened 2019-11-19
C2t3 add search event method

georginaw:
Added a search event method to the current EventModel class which queries the db with user search term on the db 'events' event name field (supports partial searching) and returns this data as an array.

### DIFF
--- a/src/Classes/Controllers/GetEventsController.php
+++ b/src/Classes/Controllers/GetEventsController.php
@@ -42,6 +42,24 @@ class GetEventsController
         if (!empty($eventSearchInput)) {
             if (strlen($eventSearchInput) < 256) {
             // run the search event function
+                $searchEvents = $this->eventModel->searchEvents($eventSearchInput);
+                if (!empty($eventSearchInput)) {
+                    $data = [
+                        'success' => true,
+                        'message' => 'Query Successful.',
+                        'data' => $eventSearchInput
+                    ];
+                    $statusCode = 200;
+                } else {
+                    $data = [
+                        'success' => true,
+                        'message' => 'There are no events.',
+                        'data' => []
+                    ];
+                    $statusCode = 200;
+                }
+                return $response->withJson($data, $statusCode);
+
             } else {
                 $data['message'] = 'Search term cannot be greater than 255 characters.';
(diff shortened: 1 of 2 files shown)


src/Classes/Controllers/GetEventsController.php
```diff
@@ -42,6 +42,24 @@ public function __invoke(Request $request, Response $response, array $args)
         if (!empty($eventSearchInput)) {
             if (strlen($eventSearchInput) < 256) {
             // run the search event function
```
georginaw: Ignore this file's changes - it is part of the next task branch and shouldn't affect this branch which is only to create the searchEvents method in the EventModel class

src/Classes/Models/EventModel.php
```diff
@@ -78,4 +78,23 @@ public function addEvent(EventEntity $newEvent):bool
...
+    {
+        $sql = 'SELECT `events`.`id`, `events`.`name`, `events`.`category`, 
+                `event_categories`.`name` AS `category_name`, `location`, `date`, `start_time`,`end_time`, `notes` 
+                FROM `events` 
+                LEFT JOIN `event_categories` ON `events`.`category` = `event_categories`.`id` 
+                WHERE `events`.`name` LIKE ? ORDER BY `date` DESC;';
+        $query = $this->db->prepare($sql);
+        $searchTerm = '%' . $searchTerm . '%';
+        $query->execute([$searchTerm]);
+        return $query->fetchAll();
```
georginaw: To test this works you may need to do a var_dump as it might not work with the current state of the GetEventsController (which we'll do in c2t4)

anton25360: What happens if the db doesn't exist? Maybe remember this for the next task. Otherwise this is all good.

avocadobelly: This is a problem for task 5.

anton25360: ok :)

anton25360: fine

anton25360: Okay, that's fine.

PR #190, opened 2019-11-20
C2t4 upadate event controller

avocadobelly:
Incorportate the event searcg method from the event model into the GetEventsController so that it will return events matching the user search. Edited front end to replace 'Current Events' title with Results when a search has been performed.


src/settings.php
```diff
@@ -18,7 +18,7 @@
 
         'db' => [
             'host' => 'mysql:host=127.0.0.1;',
-            'dbName' =>'dbname=academyPortal',
+            'dbName' =>'dbname=academyPortal2',
```
anton25360: Change db name for master compatibility

avocadobelly: changed :)

PR #192, opened 2019-11-20
Story 3 - Sort Registered applicants.

anton25360:
Uses arrows to sort the table of registered applicants by modifying the SQL query.

### DIFF
--- a/public/css/genericStyles.css
+++ b/public/css/genericStyles.css
@@ -55,19 +55,18 @@ input[type="submit"] {
     margin-bottom: 25px;
 }
 
-.arrowBtn{
+.arrowBtn {
     border: none;
 }
 
-.arrowForm{
+.arrowForm {
     display: inline-block;
-
 }
 
-.active{
-    color:#d64405;
+.active {
+    color: #d64405;
 }
 
-#applicantsPageHomeBtn{
+#applicantsPageHomeBtn {
     margin: 20px;
 }
--- a/public/css/genericStyles.css
+++ b/public/css/genericStyles.css
@@ -63,3 +63,11 @@ input[type="submit"] {
     display: inline-block;
 
 }
+
+.active{
+    color:#d64405;
+}
+
+#applicantsPageHomeBtn{
+    margin: 20px;
+}
--- a/public/css/genericStyles.css
+++ b/public/css/genericStyles.css
@@ -55,3 +55,11 @@ input[type="submit"] {
     margin-bottom: 25px;
 }
 
+.arrowBtn{
+    border: none;
+}
+
+.arrowForm{
+    display: inline-block;
+
+}
--- a/public/css/genericStyles.css
+++ b/public/css/genericStyles.css
@@ -54,3 +54,4 @@ input[type="submit"] {
 .container {
     margin-bottom: 25px;
 }
+
--- a/src/Classes/Controllers/DisplayApplicantsController.php
+++ b/src/Classes/Controllers/DisplayApplicantsController.php
@@ -36,7 +36,30 @@ class DisplayApplicantsController
      */
     public function __invoke(Request $request, Response $response, array $args)
     {
-        $args['data'] = $this->applicantModel->getAllApplicants();
+
+        $sortValue = $request->getQueryParam('sort');
+        $args['sort'] = $sortValue;
+
+        switch ($sortValue) {
+            case 'dateAsc':
+                $args['data'] = $this->applicantModel->sortApplicants('dateAsc');
+                break;
+
+            case 'dateDesc':
+                $args['data'] = $this->applicantModel->sortApplicants('dateDesc');
+                break;
+
+            case 'cohortAsc':
+                $args['data'] = $this->applicantModel->sortApplicants('cohortAsc');
+                break;
+
+            case 'cohortDesc':
+                $args['data'] = $this->applicantModel->sortApplicants('cohortDesc');
+                break;
+
+            default:
+                $args['data'] = $this->applicantModel->getAllApplicants();
+        }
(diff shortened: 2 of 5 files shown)


templates/displayApplicants.phtml
```diff
@@ -20,16 +21,29 @@
...
+                        <div>
+                            <form class="arrowForm" method="get"><button name="sort" value="dateAsc" class="arrowBtn <?php if ($sort == 'dateAsc' OR empty($sort)) {echo 'active';}?>" type="submit"><i id="arrowDateAsc" class="fas fa-arrow-up"></i></button></form>
+                            <form class="arrowForm" method="get"><button name="sort" value="dateDesc" class="arrowBtn <?php if ($sort == 'dateDesc') {echo 'active';}?>" type="submit"><i id="arrowDateDesc" class="fas fa-arrow-down"></i></button></form>
+                        </div>
+
+                    <th class="col-xs-2">Cohort date
+                        <div>
+                            <form class="arrowForm" method="get"><button name="sort" value="cohortAsc" class="arrowBtn <?php if ($sort == 'cohortAsc') {echo 'active';}?>" type="submit"><i id="arrowCohortAsc" class="fas fa-arrow-up"></i></button></form>
+                            <form class="arrowForm" method="get"><button name="sort" value="cohortDesc" class="arrowBtn <?php if ($sort == 'cohortDesc') {echo 'active';}?>" type="submit"><i id="arrowCohortDesc" class="fas fa-arrow-down"></i></button></form>
```
ashleycoles: This works well, however, there is no need to have four separate (but duplicate) forms. You can use a single form element that wraps around the table to do the same thing.

public/css/genericStyles.css
```diff
@@ -54,3 +54,20 @@ input[type="submit"] {
...
+.arrowForm{
+    display: inline-block;
+
+}
+
+.active{
+    color:#d64405;
+}
+
+#applicantsPageHomeBtn{
```
ashleycoles: Should have a space between the selector and the opening {

public/css/genericStyles.css
```diff
@@ -54,3 +54,20 @@ input[type="submit"] {
...
     margin-bottom: 25px;
 }
+
+.arrowBtn{
+    border: none;
+}
+
+.arrowForm{
+    display: inline-block;
+
```
ashleycoles: Remove empty line

public/css/genericStyles.css
```diff
@@ -54,3 +54,20 @@ input[type="submit"] {
...
+}
+
+.active{
+    color:#d64405;
```
ashleycoles: Space between property and value

src/Classes/Controllers/DisplayApplicantsController.php
```diff
@@ -36,7 +36,30 @@ public function __construct(PhpRenderer $renderer, ApplicantModel $applicantMode
      */
     public function __invoke(Request $request, Response $response, array $args)
     {
-        $args['data'] = $this->applicantModel->getAllApplicants();
+
+        $sortValue = $request->getQueryParam('sort');
+        $args['sort'] = $sortValue;
```
ashleycoles: Rather than adding your sort to the existing args array, you should create a new array that gets sent to your view (that will include your 'sort' and 'data') instead. Slims including examples take the $args array from the invoke method and pass it straight to the view, but this is bad practise in reality as it's very insecure.

src/Classes/Models/ApplicantModel.php
```diff
@@ -77,14 +77,59 @@ public function getAllApplicants()
...
+
+        $query = $this->db->prepare(
+            "SELECT `applicants`.`id`, `name`, `email`, `dateTimeAdded`, `date` 
+                      AS 'cohortDate'
+                      FROM `applicants`
+                      LEFT JOIN `cohorts` ON `applicants`.`cohortId`=`cohorts`.`id`
+                      ORDER BY $order;"
+        );
+        $query->setFetchMode(\PDO::FETCH_CLASS, 'Portal\Entities\ApplicantEntity');
+        $query->execute([$order]);
```
ashleycoles: Passing $order into this does nothing as you're not using param binding anymore.

src/settings.php
```diff
@@ -17,7 +17,7 @@
         ],
 
         'db' => [
-            'host' => 'mysql:host=db;',
+            'host' => 'mysql:host=127.0.0.1;',
```
ashleycoles: Leave these settings unchanged in master, you can just add a new commit that changes it back to db. The live site uses the config in master so it needs to remain unchanged.

templates/displayApplicants.phtml
anton25360: Wrapped the table in a form tag.

public/css/genericStyles.css
anton25360: added spaces CSS file

public/css/genericStyles.css
anton25360: Removed empty line.

public/css/genericStyles.css
anton25360: Added space.

src/Classes/Controllers/DisplayApplicantsController.php
anton25360: Created new $params array, used to store the sorting data. (Passed in the response)

src/Classes/Models/ApplicantModel.php
anton25360: Removed $order.

src/settings.php
anton25360: Changed host back to db, ready for master implementation.

PR #198, opened 2019-11-20
G2t2 Method to get Hp Ids by Event Id

sudokufan:
Wrote a method in the Event Model taking Event Id as a param, pulls out HP ids in the database that match that event id

### DIFF
--- a/src/Classes/Models/EventModel.php
+++ b/src/Classes/Models/EventModel.php
@@ -127,7 +127,7 @@ class EventModel
         }
     }
 
-    public function HpIdsByEventId($eventId): array
+    public function hpIdsByEventId($eventId): array
     {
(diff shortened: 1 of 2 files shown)


src/Classes/Models/EventModel.php
```diff
@@ -126,4 +126,13 @@ public function checklinkHP(int $hiringPartner, int $event): bool
             return false;
         }
     }
+
+    public function hpIdsByEventId($eventId): array
```
jdselby24: type hint

sudokufan: DONE

jdselby24: doc block

PR #211, opened 2019-11-21
C2 search events

avocadobelly:
Added functionality to allow users to search events including partial search. Users are then able to clear their search so that all events in db are displayed again.


public/js/eventsPage.js
```diff
@@ -10,33 +10,40 @@ const message = document.querySelector('#messages')
...
  * Runs a foreach through each event object and outputs HTML elements with event details via eventGenerator
  *
  * @param events is an array of objects which contains information about events
  */
-function displayEventsHandler(events) {
-    let eventInformation = ''
-    if (events == '') {
-        eventList.innerHTML = 'No Events Scheduled'
+function displayEventsHandler(eventsAndHiringPartners) {
+    if(eventsAndHiringPartners.events.data.length === 0) {
```
mporam: missing space:
`if (`

PR #215, opened 2019-11-21
Added removeHpfromevents  controller factory dic and routes

BuzatuCos:


src/Classes/Controllers/RemoveHiringPartnerFromEventController.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Calls a method to remove hiring partner from the event
+     *
+     * and responds with Json success message
+     *
+     * @param Request $request HTTP request
+     * @param Response $response HTTP response
+     * @param array $args
```
cwoodwarduk: Charlie will flag this as needing more description.

PR #220, opened 2020-03-30
Updated model method with new SQL statement and deleteApplicant function

mathewjmartin:
Code review needed for Task 4 - We've added a new deleteApplicant method to the ApplicantModel and updated the SQL statement in Model with a filter to make sure only it's showing not-deleted.

To test, copy the SQL statement from the code and paste into Sequel Pro and set delete status to 1 on random entries and check on the Registered Applicants page to se if they have been deleted from the table

### DIFF
--- a/src/Classes/Models/ApplicantModel.php
+++ b/src/Classes/Models/ApplicantModel.php
@@ -187,8 +187,7 @@ class ApplicantModel
         $applicantEighteenPlus,
         $applicantFinance,
         $applicantNotes
-    )
-    {
+    ) {
         return new ApplicantEntity(
             $applicantName,
             $applicantEmail,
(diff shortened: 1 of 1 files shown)


src/Classes/Models/ApplicantModel.php
```diff
@@ -201,4 +203,11 @@ public function createNewApplicant(
             $applicantNotes
         );
     }
+
+    public function deleteApplicant($id)
```
ChrisTregaskis: needs docBlocks

src/Classes/Models/ApplicantModel.php
```diff
@@ -201,4 +203,11 @@ public function createNewApplicant(
...
+    {
+        $query= $this->dbConnection->prepare( "UPDATE `applicants` SET `deleted` = '1' WHERE `id` = :id");
```
ChrisTregaskis: space after query :)

src/Classes/Models/ApplicantModel.php
robings: Doc blocks added

src/Classes/Models/ApplicantModel.php
robings: made spaces around = consistent with other functions in file.

PR #228, opened 2020-04-01
S2 t4 a stages entity

daniellaskinner:
please run php unit test on StagesEntityTest.php

### DIFF
--- a/src/Classes/Entities/StageEntity.php
+++ b/src/Classes/Entities/StageEntity.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace Portal\Entities;
+
+class StageEntity extends ValidationEntity implements \JsonSerializable
+{
+    protected $id;
+    protected $title;
+    protected $order;
+
+    public function __construct(string $id, string $title, string $order)
+    {
+        $this->id = $id;
+        $this->title = $title;
+        $this->order = $order;
+        $this->sanitiseData();
+    }
+
+    /**
+     * Returns private properties from object.
+     *
+     * @return array|mixed
+     */
+    public function jsonSerialize()
+    {
+
+        return [
+            'id' => $this->id,
+            'title' => $this->title,
+            'order' => $this->order
+        ];
+    }
+
+    /**
+     * Will sanitise all the fields for a stage.
+     */
+    private function sanitiseData()
+    {
+        $this->id = (int) $this->id;
+        $this->title = self::sanitiseString($this->title);
+        $this->order = (int) $this->order;
+    }
+
+    /**
+     * Get's stage Id.
+     *
+     * @return int, returns the stage Id field.
+     */
+    public function getStageId(): int
+    {
+        return $this->id;
+    }
+
+    /**
+     * Get's stage title.
+     *
+     * @return string, returns the stage title field.
+     */
+    public function getStageTitle(): string
+    {
+        return $this->title;
+    }
+
+    /**
+     * Get's stage order.
+     *
+     * @return int, returns the stage order field.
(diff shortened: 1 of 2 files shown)


src/Classes/Entities/StageEntity.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Portal\Entities;
+
+class StageEntity extends ValidationEntity implements \JsonSerializable
+{
+    protected $id;
+    protected $title;
+    protected $order;
+
+    public function __construct(string $id = null, string $title = null, string $order = null)
```
ChrisTregaskis, line 11: docBlock

src/Classes/Entities/StageEntity.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        $this->id = ($this->id ?? $id);
```
ChrisTregaskis, line 13: is the null coalescing operator really necessary lol? Happy for it to stay of-course. Just takes a few more looks to understand whats happening.

src/Classes/Entities/StageEntity.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Will sanitise all the fields for a stage.
+     */
+    private function sanitiseData()
+    {
+        $this->id = (int) $this->id;
+        $this->title = self::sanitiseString($this->title);
+        $this->order = (int) $this->order;
+    }
```
ChrisTregaskis, line 42: does this work? Does it need to return anything?

tests/Entities/StageEntityTest.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Tests\Entities;
+
+use Portal\Entities\StageEntity;
+use \Exception;
+use PHPUnit\Framework\TestCase;
+use TypeError;
+
+class StageEntityTest extends TestCase
+{
+    public function testNewStageEntitySuccess()
```
ChrisTregaskis, line 12: docBlocks for each test

tests/Entities/StageEntityTest.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Tests\Entities;
+
+use Portal\Entities\StageEntity;
+use \Exception;
```
ChrisTregaskis, line 6: I've not seen exception being used, is it necessary in this test file?

tests/Entities/StageEntityTest.php
```diff
@@ -0,0 +1,54 @@
...
+use PHPUnit\Framework\TestCase;
+use TypeError;
```
ChrisTregaskis, line 8: I've not seen TypeError being used, is it necessary in this test file?

src/Classes/Entities/StageEntity.php
robings: No other entity has a doc block on the constructor

src/Classes/Entities/StageEntity.php
robings: Yes, we think it is because we will be using this entity in the same way as other entities in the application that have this operator.

src/Classes/Entities/StageEntity.php
robings: yes, it is replacing the contents of the parameters with santized versions of themselves.

tests/Entities/StageEntityTest.php
robings: There aren't any in other tests, however, I think in this case they should be, so have added them.

PR #233, opened 2020-04-02
S2 t5 create stage

paulms1:


src/Classes/Entities/StageEntity.php
```diff
@@ -17,6 +17,7 @@ public function __construct($title = null, $order = null)
         $this->order = ($this->order ?? $order);
         $this->deleted = 0;
 
+
```
daniellaskinner, line 20: remove this extra line.. not sure if you guys added it or someone else..

PR #239, opened 2020-05-15
General tidies, mostly validation and best practise

mporam:


src/Controllers/EditStageController.php
```diff
@@ -2,14 +2,14 @@
 
 namespace Portal\Controllers;
 
-use Portal\Entities\StageEntity;
-use Portal\Entities\ValidationEntity;
+use Portal\Abstracts\Controller;
+use Portal\Validators\StageValidator;
+use Portal\Validators\StringValidator;
```
ashleycoles, line 7: can ditch this, it's not used

PR #242, opened 2020-05-18
Added validation for add new applicant page. Removed character limit …

danicatt:
…on text areas.

### DIFF
--- a/public/js/saveApplicant.js
+++ b/public/js/saveApplicant.js
@@ -44,22 +44,13 @@ let makeApiRequest = async (data) => {
 }
 
 let validateFormInputs = (data) => {
-    let validate = []
+    let validate = [];
 
-    validate.push(evalLength(data.name))
-    validate.push(evalLength(data.email))
-    validate.push(evalLength(data.phoneNumber))
-    validate.push(evalLength(data.whyDev))
-    validate.push(evalLength(data.codeExperience))
-    validate.push(evalLength(data.hearAboutId))
+    validate.push(isName(data.name));
+    validate.push(isEmail(data.email));
+    validate.push(isPhoneNumber(data.phoneNumber));
+    validate.push(evalLength(data.whyDev));
+    validate.push(evalLength(data.codeExperience));
 
-    return validate.includes(false) ? false : true
-}
-
-function evalLength(data) {
-    if(data.length >= 1 && data.length <= 255){
-        return true
-    } else {
-        return false
-    }
-}
+    return validate.includes(false) ? false : true;
+};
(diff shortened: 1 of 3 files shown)


public/js/saveApplicant.js
```diff
@@ -44,22 +44,13 @@ let makeApiRequest = async (data) => {
...
 let validateFormInputs = (data) => {
-    let validate = []
+    let validate = [];
 
-    validate.push(evalLength(data.name))
-    validate.push(evalLength(data.email))
-    validate.push(evalLength(data.phoneNumber))
-    validate.push(evalLength(data.whyDev))
-    validate.push(evalLength(data.codeExperience))
-    validate.push(evalLength(data.hearAboutId))
```
bradleypurdue, line 54: Is there a reason why we don't have this anymore?

public/js/validations.js
```diff
@@ -8,16 +8,25 @@ function isDate(date) {
...
         return true
     } else {
         return false
     }
 }
 
+function isName(name) {
+    let pattern = /^[A-Za-z][A-Za-z\'\-]+([\ A-Za-z][A-Za-z\'\-]+)*/;
+    if (pattern.test(name)) {
+        return true
```
bradleypurdue, line 22: semi colon :)

public/js/validations.js
```diff
@@ -8,16 +8,25 @@ function isDate(date) {
...
+    } else {
+        return false
+    }
```
bradleypurdue, line 25: Since if the condition is met on the first it will `return` this `else` is not needed, you can just have:
```js
if (pattern.test(name)) {
    return true;
}

return false;
```

public/js/validations.js
```diff
@@ -45,11 +54,19 @@ function isUrl(url) {
...
+    let regEx = /\b((?:(?:gir)|(?:[a-pr-uwyz])(?:(?:[0-9](?:[a-hjkpstuw]|[0-9])?)|(?:[a-hk-y][0-9](?:[0-9]|[abehmnprv-y])?)))) ?([0-9][abd-hjlnp-uw-z]{2})\b/ig;
     if (regEx.test(postcode)) {
         return true
     } else {
         return false
     }
 }
 
+function evalLength(data) {
+    if(data.length >= 1) {
```
bradleypurdue, line 66: Can we add a space between`if` and `(` please.

public/js/validations.js
```diff
@@ -45,11 +54,19 @@ function isUrl(url) {
...
+        return true
+    } else {
```
bradleypurdue, line 68: Same thing regarding the else here.

public/js/saveApplicant.js
danicatt: We took it out as it wasn't doing anything. hearAboutId had been changed to a drop-down.

public/js/validations.js
danicatt: added :)

PR #243, opened 2020-05-18
Added validation to text area fields, limiting to 10000 chars.

danicatt:


public/js/saveApplicant.js
```diff
@@ -46,11 +46,11 @@ let makeApiRequest = async (data) => {
 let validateFormInputs = (data) => {
     let validate = [];
 
-    validate.push(isName(data.name) && charsMaxLength(data.name));
+    validate.push(isName(data.name) && nameMaxLength(data.name));
     validate.push(isEmail(data.email));
```
turnerror, line 50: E-mail max length is also 255 varchar

public/js/saveApplicant.js
```diff
@@ -46,11 +46,11 @@ let makeApiRequest = async (data) => {
...
     validate.push(isPhoneNumber(data.phoneNumber));
-    validate.push(isPresent(data.whyDev));
-    validate.push(isPresent(data.codeExperience));
+    validate.push(isPresent(data.whyDev) && textAreaMaxLength(data.whyDev));
+    validate.push(isPresent(data.codeExperience) && textAreaMaxLength(data.codeExperience));
```
turnerror, line 54: This is stupid haha, but the Mike and Charlie's notes bit should probably also be capped at 10k text haha.

PR #244, opened 2020-05-18
aTask1c, Added validation to application entity, Removed useless method from ap…

turnerror:
…plicant model and changed saveapplicant controller to apply changes.

### DIFF
--- a/src/Controllers/SaveApplicantController.php
+++ b/src/Controllers/SaveApplicantController.php
@@ -40,7 +40,7 @@ class SaveApplicantController extends Controller
     {
         if ($_SESSION['loggedIn'] === true) {
             $data = ['success' => false, 'msg' => 'Application not saved', 'data' => []];
-            $statusCode = 401;
+            $statusCode = 500;
 
             $newApplicationData = $request->getParsedBody();
 
--- a/src/Controllers/SaveApplicantController.php
+++ b/src/Controllers/SaveApplicantController.php
@@ -2,8 +2,11 @@
 
 namespace Portal\Controllers;
 
+use Exception;
 use Portal\Abstracts\Controller;
+use Portal\Entities\ApplicantEntity;
 use Portal\Models\ApplicantModel;
+use Portal\Validators\NewApplicantValidator;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Psr\Http\Message\ResponseInterface as Response;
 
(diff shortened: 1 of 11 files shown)


src/Controllers/SaveApplicantController.php
```diff
@@ -31,40 +34,46 @@ public function __construct(ApplicantModel $applicantModel)
      * @param array $args The arguments array
      *
      * @return Response, will return the data from successfulRegister and the statusCode, via Json.
+     * @throws Exception
      */
-    public function __invoke(Request $request, Response $response, Array $args)
+    public function __invoke(Request $request, Response $response, array $args)
```
amandahrdmn, line 39: why change the Array to array?

src/Validators/PhoneNumberValidator.php
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Make sure the phone Number is valid
+     *
+     * @param string $phoneNumber
+     * @return string|null
+     * @throws Exception
+     */
+    public static function validatePhoneNumber(string $phoneNumber)
+    {
+        if ($phoneNumber == '') {
```
amandahrdmn, line 22: don't you want to throw the 400 error if you don't have a phone number?

tests/Validators/PhoneNumberValidatorTest.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Tests\Validators;
+
+use Portal\Validators\PhoneNumberValidator;
+
+class PhoneNumberValidatorTest extends \PHPUnit\Framework\TestCase
+{
+
+    public function testValidatePhoneNumberSuccess()
+    {
+        $numbers = ['786780670434', '+31 166 094'];
```
amandahrdmn, line 12: need to check what happens if you don't input a phone number

amandahrdmn:
just realised that StringValidatorTest also needs checks for empty strings in the validateLength function

src/Controllers/SaveApplicantController.php
turnerror: psr standard

turnerror:
> just realised that StringValidatorTest also needs checks for empty strings in the validateLength function

The validate length function should be able to be empty. That's why there's 2, validateLength and validateExistsAndLength

PR #255, opened 2020-05-20
Story4

JamesWaterhouse:


public/js/stageReorder.js
```diff
@@ -16,9 +16,10 @@ Sortable.create(stageTable, {
             stageTableRows.forEach((tableRow)=>{
                 let stageEntity = {};
                 stageEntity.id = tableRow.dataset.id;
-                stageEntity.title = tableRow.children[0].children[0].textContent;
+                stageEntity.title = tableRow.children[1].children[0].textContent;
                 stageEntity.order = (stageOrder.indexOf(tableRow.dataset.id) + 1);
                 requestData.data.push(stageEntity);
+                location.reload(true);
```
ashleycoles, line 22: No need for the true in here, just location.reload() will do it

PR #259, opened 2020-05-20
S2t4 display options with stages

richherbs:


public/js/stagesPage.js
```diff
@@ -4,6 +4,64 @@ const editResponse = document.getElementById('editResponse');
...
+    optionAddSubmit.addEventListener('click', async (e) => {
+        e.preventDefault()
+        let inputs = e.target.parentElement.querySelectorAll('input[type="text"]')
+        let stageId = e.target.dataset.stageid
+
+        let data = {
+            "data" : []
+        };
+        inputs.forEach((input) => {
+            data.data.push({
```
hannahgolightly, line 56: bit confusing calling them both data- does it need to be?

src/ViewHelpers/StagesPageViewHelper.php
```diff
@@ -18,18 +19,54 @@ public static function displayStages($stages)
...
+                        $result .= '</div>';
+                    }
+                    $result .= '<div class="optionContainer">';
+                    $result .= '<form data-id="" class="optionAddForm ">';
+                }
+                $result .= '<input type="text" class="optionAddTitle col-xs-10" ';
+                $result .= 'placeholder="Type the name of your new option"/>';
+                $result .= '<input type="submit" class="optionAddSubmit btn-success" data-stageid="';
+                $result .= $stage->getStageId().'" value="Submit">';
+                $result .= '</form>';
```
hannahgolightly, line 60: maybe could combine some of these shorter lines? only a thought, just there's a lot of $result .=

tests/ViewHelpers/StagesPageViewHelperTest.php
```diff
@@ -15,22 +16,49 @@ class StagesPageViewHelperTest extends TestCase
...
+        $expected .= '<td class="col-xs-2 text-center"><a class="toggleEditForm">Edit</a></td>';
+        $expected .= '<td class="col-xs-2 text-center"><a data-id="1" class="text-danger delete">Delete</a></td>';
+        $expected .= '<td class="col-xs-2 text-center"><a class="toggleEditOptions" data-stageId="1">Options</a></td>';
         $expected .= '</tr>';
-        $entityMock = $this->createMock(StageEntity::class);
-        $entityMock->method('getStageId')->willReturn('1');
-        $entityMock->method('getStageTitle')->willReturn('Stage 1 test');
 
-        $input =  [$entityMock];
+        $objectEntityMock = $this->createMock(OptionsEntity::class);
```
hannahgolightly, line 52: do you mean optionEntityMock here?

src/ViewHelpers/StagesPageViewHelper.php
richherbs: unfortunately this is to keep below the 120 char limit of circle-ci

PR #263, opened 2020-05-21
aTask3c

bradleypurdue:
https://trello.com/c/9ZR188RL/229-3c-update-view-to-accommodate-dropdown-with-view-helper

### DIFF
--- a/public/js/populateDropdowns.js
+++ b/public/js/populateDropdowns.js
@@ -27,7 +27,7 @@ let outputCohorts = async () => {
         cohortOptions += `value="${item.id}">${item.date}</option>`;
     });
 
-    element.innerHTML = cohortOptions;
+    element.innerHTML += cohortOptions;
 };
 
 let outputHearAbout = async () => {
@@ -45,5 +45,3 @@ let outputHearAbout = async () => {
 
     element.innerHTML = hearAboutOptions;
 };
-
-
--- a/public/js/populateDropdowns.js
+++ b/public/js/populateDropdowns.js
@@ -33,6 +33,3 @@ let outputHearAbout = async () => {
 
     hearAbout.innerHTML = hearAboutOptions.join('')
 }
-
-outputCohorts()
-outputHearAbout()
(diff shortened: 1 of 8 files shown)


templates/displayApplicants.phtml
```diff
@@ -10,15 +10,22 @@
...
     <script defer src="./js/deleteApplicant.js"></script>
+    <script defer src="./js/displayApplicants.js"></script>
 </head>
 <body>
 <div class="container">
     <div class="row">
         <div class="col-xs-12 registeredApplicants">
             <h1>Registered Applicants</h1>
+            <div class="col-4"><label for="cohorts">Select Cohort</label>
+            </div>
```
turnerror, line 24: This needs to be removed.

public/js/populateDropdowns.js
```diff
@@ -27,7 +27,7 @@ let outputCohorts = async () => {
         cohortOptions += `value="${item.id}">${item.date}</option>`;
     });
 
-    element.innerHTML = cohortOptions;
+    element.innerHTML += cohortOptions;
 };
 
 let outputHearAbout = async () => {
```
alexjennings56, line 33: can this be a const? sorry just can't see it being re-assigned unless this is the reassignment? Just ignore if i'm wrong.

templates/displayApplicants.phtml
```diff
@@ -10,15 +10,23 @@
...
     <div class="row">
         <div class="col-xs-12 registeredApplicants">
             <h1>Registered Applicants</h1>
             <form class="arrowForm" method="get">
+				<label for="cohorts">
+					Select Cohort
+					<select id="cohorts" class="submitApplicant form-control" name="cohortId">
+						<option value="all">All</option>
+					</select>
+				</label>
```
alexjennings56, line 29: indentation seems a bit off here? unless this is the correct way?

bradleypurdue: I think this is GitHub being funny, looks fine in PHPStorm.

public/js/populateDropdowns.js
bradleypurdue: This isn't our code so no clue.

PR #264, opened 2020-05-21
S2t6 delete logic

HiredMussel:


templates/stagesPage.phtml
```diff
@@ -36,5 +36,15 @@
         </div>
     </div>
 </div>
+<div class="modalContainer hidden">
+    <div class="container">
+        <div class="body">
+            <b>Warning: Deleting this option will delete all options from this stage.
```
isabellaaston, line 42: is b tag appropriate, looks like the b tag is only to be used as a last resort? (according to w3schools lol)

PR #265, opened 2020-05-21
atask3d

SamCrudge:
Hidden Cohort sort buttons when needed.


templates/displayApplicants.phtml
```diff
@@ -38,7 +38,7 @@
                             </div>
 
                         <th class="col-xs-2">Cohort date
```
alexjennings56, line 40: can't see the closing tag for this but it's probs under the </div> so i'm sure it's fine.

PR #272, opened 2020-05-21
A task2 ... update editApplicant controller so applicant can be edited.

turnerror:


public/js/saveApplicant.js
```diff
@@ -80,12 +86,15 @@ let makeApiRequest = async (data) => {
         body: JSON.stringify(data)
     })
         .then(response => {
+            console.log(response);
```
turnerror, line 89: nice one christian

public/js/saveApplicant.js
```diff
@@ -58,6 +58,11 @@ let errorMessage = (validationType) => {
 let getCompletedFormData = () => {
     const formData = document.querySelectorAll(".submitApplicant");
     let data = {};
+    const queryString = window.location.search;
+    const urlParams = new URLSearchParams(queryString);
+    const id = urlParams.get('id');
+    data['id'] = id;
```
bradleypurdue, line 64: Does it not make more sense just to do `data['id'] = urlParams.get('id');`?

public/js/saveApplicant.js
bradleypurdue: nice one christian

public/js/saveApplicant.js
```diff
@@ -69,8 +74,9 @@ let getCompletedFormData = () => {
     return data;
 };
 
-let makeApiRequest = async (data) => {
-    return fetch('./api/saveApplicant', {
+let makeApiRequest = async (data, type) => {
+    let path = './api/' + (type === '/addapplicant' ? 'saveApplicant' : 'editApplicant');
```
bradleypurdue, line 78: `const`?

src/Models/ApplicantModel.php
```diff
@@ -151,4 +151,47 @@ public function deleteApplicant($id)
...
+                            `phoneNumber` = :phoneNumber,
+                            `cohortId` = :cohortId,
+                            `whyDev` = :whyDev,
+                            `codeExperience` = :codeExperience,
+                            `hearAboutId` = :hearAboutId,
+                            `eligible` = :eligible,
+                            `eighteenPlus` = :eighteenPlus,
+                            `finance` = :finance,
+                            `notes` = :notes
+
```
SamCrudge, line 176: white space?

public/js/saveApplicant.js
SamCrudge: nice one christian

PR #275, opened 2020-06-24
Add new fields to applicants

mporam:


mporam:
changed merged into team-picker branch

PR #288, opened 2020-11-24
S1 stage filter

jjtapley:


src/Models/ApplicantModel.php
```diff
@@ -76,53 +76,32 @@ public function storeApplicant(ApplicantEntityInterface $applicant)
...
-        $query = $this->db->prepare($stmt);
-        $query->setFetchMode(\PDO::FETCH_CLASS, BaseApplicantEntity::class);
-
-        $query->execute();
-
-        return $query->fetchAll();
-    }
-
-    /**
-     * Gets a sorted list of appliciants assigned to a specific cohort.
```
ashleycoles, line 105: Don't forget to update the main description in this docblock

public/js/displayApplicants.js
```diff
@@ -1,24 +1,77 @@
...
+      .querySelector('.stagesSort')
+      .classList.add('hidden');
+  } else {
+    document
+      .querySelector("[value='all']")
+      .setAttribute('selected', 'true');
+    document
+      .querySelector('.stagesSort')
+      .classList.remove('hidden');
+  }
```
ashleycoles, line 76: Is this doing anything? It looks to me like it's used for the cohorts, because as well as filtering by cohort you can also sort by cohort, but this is not the case for stages

ashleycoles (changes requested):

public/js/displayApplicants.js
```diff
@@ -1,24 +1,61 @@
...
+    document
+      .querySelector('.stagesSort')
+      .classList.add('hidden');
+  } else {
+    document
+      .querySelector("[value='all']")
+      .setAttribute('selected', 'true');
+    document
+      .querySelector('.stagesSort')
+      .classList.remove('hidden');
```
ashleycoles, line 59: This can come out, there is no stagesSort

public/js/displayApplicants.js
```diff
@@ -1,24 +1,61 @@
...
+    form.submit();
+  });
+
+  if (stage !== 'all') {
+    document
+      .querySelector(`[value='${stage}']`)
+      .setAttribute('selected', 'true');
+    document
+      .querySelector('.stagesSort')
+      .classList.add('hidden');
```
ashleycoles, line 52: This can come out, there is no stagesSort

ashleycoles (changes requested):

ashleycoles (approved):

PR #290, opened 2020-11-25
Mp story1 t6

JoshThomasField:


src/Controllers/API/EditApplicantController.php
```diff
@@ -61,13 +61,21 @@ public function __invoke(Request $request, Response $response, array $args)
...
                     $newApplicationData['taster'],
-                    $newApplicationData['tasterAttendance']
+                    $newApplicationData['tasterAttendance'],
+                    null,
+                    $newApplicationData['stageId'],
+                    $newApplicationData['optionId']
                 );
+                //If applicant is being moved to a stage that makes them a student a row in the
+                // applicant_additional table needs to be created. If it doesn't already exist.
+
```
elstevens23, line 71: Is there a need for two lines breaks here? If not, please remove :)

elstevens23: Also on lines 69 and 70, if the comments do not relate to the code above, add a line break there too (between lines 69 and 70).

src/Entities/CompleteApplicantEntity.php
```diff
@@ -27,7 +27,10 @@ class CompleteApplicantEntity extends ApplicantEntity implements \JsonSerializab
     protected $laptopNum; // int
     protected $taster; // date string
     protected $tasterAttendance; // bool
+    protected $stageId; // int
     protected $team; // int
+    protected $stageID; //
```
elstevens23, line 32: Please remove these two forward slashes as there is no commend here :)

elstevens23 (changes requested):
There are just a couple of syntactical changes 👍

elstevens23 (approved):
💯

PR #295, opened 2020-11-25
S2 paginate applicant list

jjtapley:


app/routes.php
```diff
@@ -9,7 +9,7 @@
     $app->get('/admin', 'AdminController');
     $app->get('/register', 'RegisterController');
     $app->get('/addapplicant', 'addApplicantController');
-    $app->get('/applicants', 'DisplayApplicantsController');
+    $app->get('/applicants[/{page}]', 'DisplayApplicantsController');
```
ashleycoles, line 12: the optional/dynamic portion of the URL doesn't get used, and if you manually go to /applicants/2 for example the page is just broken.

Suggest removing this and just making the route for /applicants

templates/applicants.phtml
```diff
@@ -53,7 +53,27 @@
...
+                            <a class="page-link" href="/applicants?page=<?= $i; ?>"> <?= $i; ?> </a>
+                        </li>
+                    <?php endfor; ?>
+
+                    <li class="page-item <?php if($page >= $count) { echo 'disabled'; } ?>">
+                        <a class="page-link"
+                           href="<?php if($page >= $count){ echo '#'; } else {echo "?page=". ($page+1); } ?>">Next</a>
+                    </li>
+                </ul>
+            </nav>
```
ashleycoles, line 74: Now that we've added the pagination for applicants, we may well want to add the same to the other tables in the portal. Therefore it would be quite nice if this code were re-usable. 

Would suggest putting the pagination display into a generic pagination view helper that could then be used elsewhere if needed.

src/Models/ApplicantModel.php
```diff
@@ -94,14 +120,18 @@ public function getApplicants(string $stageId = '%', string $cohortId = '%', str
                       LEFT JOIN `options` ON `applicants`.`stageOptionId` = `options`.`id`
                       WHERE `applicants`.`deleted` = '0'
                       AND `applicants`.`cohortId` like :cohortId
-                      AND `applicants`.`stageId` like :stageId; ";
+                      AND `applicants`.`stageId` like :stageId
+                      LIMIT :offsets, :numberPerPage;";
 
-        $stmt .= $this->sortingQuery($sortingQuery);
 
+        $stmt .= $this->sortingQuery($sortingQuery);
```
ashleycoles, line 127: The sorting no longer seems to work! Neither sorting by date or cohort!

ashleycoles (changes requested):

src/ViewHelpers/PaginationViewHelper.php
```diff
@@ -0,0 +1,63 @@
...
+     * @return string
+     */
+    public static function applicantsPagination($page, $count)
+    {
+        $next = (strval($page+1));
+        $prev = (strval($page-1));
+
+        $applicantPagination = '';
+
+        $applicantPagination .= '<nav aria-label="Page navigation mt-5">';
```
ashleycoles, line 24: Incorrect use of the aria-label, is mt-5 not a bootstrap class?

src/ViewHelpers/PaginationViewHelper.php
```diff
@@ -0,0 +1,63 @@
...
+        $applicantPagination .= '<ul class="pagination justify-content-center">';
+        $applicantPagination .= '<li class="page-item ';
+        if ($page <= 1) {
+            $applicantPagination .= 'disabled';
```
ashleycoles, line 28: Can you add some CSS for this disabled class that sets pointer-events: none; on the a tag? This way the user wont even be able to click on it.

src/ViewHelpers/PaginationViewHelper.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+        $applicantPagination .= '"><a class="page-link" href="';
+        if ($page <= 1) {
+            $applicantPagination .= '#';
+        } else {
+            $applicantPagination .= '/applicants?page=' . $prev;
```
ashleycoles, line 34: If you remove '/applicants' from each of these links, it should still work and would mean that this view helper could be used universally across the app whenever we want some pagination

src/ViewHelpers/PaginationViewHelper.php
```diff
@@ -0,0 +1,63 @@
...
+        $applicantPagination .= '">Previous</a></li>';
+
+        for ($i = 1; $i <= $count; $i++) :
+            $applicantPagination .= '<li class="page-item ';
+            if ($page == $i) {
+                $applicantPagination .= 'active';
+            }
+            $applicantPagination .= '">';
+            $applicantPagination .= '<a class="page-link" href="/applicants?page=' . $i . '">' . $i . '</a></li>';
+        endfor;
```
ashleycoles, line 45: Switch to the normal for syntax of for () {} rather than using endfor;

we only really use that alternate syntax when we're nesting the loop amongst a bunch of HTML because it makes it slightly easier to read.

tests/ViewHelpers/PaginationViewHelperTest.php
```diff
@@ -0,0 +1,34 @@
...
+use Tests\TestCase;
+
+class PaginationViewHelperTest extends TestCase
+{
+    public function testSuccessPaginationViewHelper()
+    {
+        $expected = '<nav aria-label="Page navigation mt-5">
+                        <ul class="pagination justify-content-center">
+                            <li class="page-item disabled"><a class="page-link" href="#">Previous</a></li>
+                            <li class="page-item active"><a class="page-link" href="/applicants?page=1">1</a></li>
```
ashleycoles, line 15: This will need to be updated once you take out '/applicants'

src/ViewHelpers/PaginationViewHelper.php
ashleycoles: Then you could change the method name so it's more generic

ashleycoles (changes requested):

ashleycoles (approved):

PR #300, opened 2020-11-26
Mp s1 t4 build pop up

elstevens23:
When clicking the 'next stage' button on the Applicants table, should be moved to next stage or given a pop up to choose from stage options. When the applicant is on the final stage, the next stage button will disappear.


src/Controllers/DisplayNewStageOptionsController.php
```diff
@@ -0,0 +1,47 @@
...
+    public function __invoke(Request $request, Response $response, array $args)
+    {
+        if ($_SESSION['loggedIn'] === true) {
+            $args['eventType'] = 'Upcoming';
+            return $this->renderer->render($response, 'events.phtml', $args);
+        } else {
+            return $response->withHeader('Location', './');
+        }
+    }
+}
```
JoshThomasField, line 47: This file appears to be redundant?

public/js/nextStageOptionsPage.js
```diff
@@ -0,0 +1,54 @@
...
+//     }
+// })
+//
+// document.querySelector('#clear-search').addEventListener('click', function(e) {
+//     e.preventDefault()
+//     if (!window.location.href.includes('#events-list')) {
+//         window.location.href += '#events-list'
+//     }
+//     location.reload()
+// })
```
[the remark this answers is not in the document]
elstevens23: We deleted the whole file as it wasn't being called and so wasn't necessary.

src/Controllers/DisplayNewStageOptionsController.php
elstevens23: Agreed. It doesn't do anything. We deleted this file also.

JoshThomasField (approved):

PR #303, opened 2020-11-26
Mp s1 unittesting

marcusxavier85:


tests/Entities/ApplicantEntityTest.php
```diff
@@ -20,7 +20,7 @@ private function createApplicant()
             '1',
             '1',
             '1',
-            'notes'
+            'notes',
```
jennymegan, line 23: does this need a comma on the last one?

marcusxavier85: no

PR #312, opened 2021-04-08
Story2task1 fix dropdown

HarryCusworth:
Damien and I fixed the problem but made an indenting problem, this fixed now too.
OUTSTANDING: some of the entries in the database table for stages or options don't make sense, might need Mike O to clarify what really needs to be there, it shouldn't affect the code.


src/ViewHelpers/ApplicantFormViewHelper.php
```diff
@@ -19,20 +19,30 @@ public static function stagesDropdown($stages, $stageOptions)
                 }
             }
             if ($stageHasOptions) {
-                $string .= '<optgroup label="' . $stage['title'] . '">';
+                $string .= '<optgroup label="' . $stage['title'] . '" >';
```
Emily774: Not sure about the function of this added space?

Emily774 (approved):

PR #315, opened 2021-04-08
Story2 sloths

Emily774:
Fixed non-existant feild in drop down list

### DIFF
--- a/tests/ViewHelpers/ApplicantFormViewHelperTest.php
+++ b/tests/ViewHelpers/ApplicantFormViewHelperTest.php
@@ -18,13 +18,15 @@ class ApplicantFormViewHelperTest extends TestCase
             'stringThree' => ['id' => 1, 'option' => 'dummy-option', 'stageId' => 1]
         ];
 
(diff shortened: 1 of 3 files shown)


tests/ViewHelpers/ApplicantFormViewHelperTest.php
```diff
@@ -18,13 +18,15 @@ public function testSuccessApplicantForm()
             'stringThree' => ['id' => 1, 'option' => 'dummy-option', 'stageId' => 1]
         ];
 
+        $currentStage = ['dummy-selected'];
```
norayna, line 21: need to change this to match one of the stages that are defined here, currently no stage has this name

norayna (changes requested):
everything all good otherwise!

tests/ViewHelpers/ApplicantFormViewHelperTest.php
HarryCusworth: its done

norayna (approved):

PR #323, opened 2021-05-24
Added tasterId field to applicants_additional table

MattW-R:


db/academyPortal.sql
```diff
@@ -21,446 +21,57 @@ SET NAMES utf8mb4;
 /*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
 
 
-# Dump of table applicants
```
Zedonk1995, line 24: I think you've deleted all of the tables except applicants_additional.  Could you please add the other tables back in?

Zedonk1995 (changes requested):

Zedonk1995 (approved):

PR #329, opened 2021-05-24
S1t3

isabellecochrane:
model getters for new fieldnames and updatd query

### DIFF
--- a/src/Entities/CompleteApplicantEntity.php
+++ b/src/Entities/CompleteApplicantEntity.php
@@ -36,6 +36,20 @@ class CompleteApplicantEntity extends ApplicantEntity implements \JsonSerializab
     protected $additionalNotes;
     protected $chosenCourseId;
     protected $chosenCourseDate;
+    protected $attitude;
+    protected $averageScore;
+    protected $fee;
+    protected $signedTerms;
+    protected $signedDiversitech;
+    protected $inductionEmailSent;
+    protected $signedNDA;
+    protected $checkedID;
+    protected $dpName;
+    protected $dpPhoto;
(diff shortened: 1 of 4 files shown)


src/Entities/CompleteApplicantEntity.php
```diff
@@ -36,6 +36,20 @@ class CompleteApplicantEntity extends ApplicantEntity implements \JsonSerializab
     protected $additionalNotes;
     protected $chosenCourseId;
     protected $chosenCourseDate;
+    protected $attitude;
```
imypullan, line 39: The others aren't typehinted so I don't know why these would be but should these be typehinted?

imypullan: wait is it because they return mixed?

imypullan: but why do they return mixed?

src/Models/ApplicantModel.php
```diff
@@ -219,7 +226,7 @@ public function getApplicantById($id)
                             ON `applicants`.`stageId` = `stages`.`id`
                         LEFT JOIN `options` 
                             ON `applicants`.`stageOptionId` = `options`.`id`
-                        WHERE `applicants`.`id`= :id;"
+                        WHERE `applicants`.`id`= ':id';"
```
imypullan, line 229: A statement rather than a comment - what a chunky query!!!!!!

src/Entities/CompleteApplicantEntity.php
isabellecochrane: shhhh

PR #337, opened 2021-05-25
Story9

TobySTW:
s9t2 - added profile link HTML to bottom of applicant details
s9t3 - added working hyperlink to the profile link (takes to dynamic profile page)
s9t4 - added clipboard icon which copies the link to the clipboard + shows when it has been clicked

### DIFF
--- a/public/js/applicantModal.js
+++ b/public/js/applicantModal.js
@@ -52,7 +52,7 @@ function copyToClipboard(element) {
 $(document).ready(function(){
     $(".myBtn").click(function(){
         var url = './api/getApplicant/' + this.dataset.id
-        let studentUrl = 'http://localhost:8080/public/'+this.dataset.id
+        var studentUrl = 'http://localhost:8080/public/'+this.dataset.id
         fetch(url)
             .then(
                 function(response) {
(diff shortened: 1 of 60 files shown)


templates/partials/applicantInfoModal.phtml
```diff
@@ -69,6 +69,11 @@
                         <p class="detail">Portfolio Url: <span id="portfolioUrl"></span></p>
                         <p class="detail">Plesk Hosting Url: <span id="pleskHostingUrl"></span></p>
                         <p class="detail">Github Education Link: <span id="githubEducationLink"></span></p>
+                        <p class="detail">Profile Link: <a id="userProfileLink" href="userProfileLink"></a> <button class="clipboard" onclick="copyToClipboard('#userProfileLink')"> <i class="glyphicon glyphicon-copy"></i>
```
ashleycoles, line 72: Can you make the link open in a new tab please

ashleycoles: Also inline event listeners are the devils work - please remove the onclick here and add an event listener in the JS instead

public/js/applicantModal.js
```diff
@@ -40,11 +40,19 @@ function addCurrency(number) {
...
+    $temp.val($(element).text()).select();
+    document.execCommand("copy");
+    $temp.remove();
+    document.querySelector("button.clipboard").innerText = 'Copied';
+}
 
 $(document).ready(function(){
     $(".myBtn").click(function(){
         var url = './api/getApplicant/' + this.dataset.id
+        let studentUrl = 'http://localhost:8080/public/'+this.dataset.id
```
ashleycoles, line 55: Switch to a var for consistency sake - even though let is better

ashleycoles (changes requested):

ashleycoles (dismissed):

public/js/applicantModal.js
mporam: disagree, swap them both to lets, better than they found it over consistency

mporam (approved):

PR #349, opened 2021-05-26
S4t4

nickclifford904:
Added more helpful error messages for FE form and checked consistency again on the BE validation.


dotNATE (commented):
All seems sound to me. Good to go!

PR #351, opened 2021-05-26
Story1

isabellecochrane:


templates/applicantForm.phtml
```diff
@@ -57,6 +57,9 @@
                     <select id="cohorts" class="submitApplicant form-control" name="cohortId"  data-selected="<?php echo $applicant->getCohortId();?>" required>
                     </select>
 
+                    <label for="backgroundInfoId">Background Info</label>
+                    <select id="backgroundInfo" class="submitApplicant form-control" name="backgroundInfoId" required></select>
```
ashleycoles, line 61: This field does not show the current value - other dropdown fields have a default option which is whatever is in the db for that applicant, can you set this up to do the same please.

To reproduce just edit an applicant, change their background to school leaver and save, then refresh. The background field will just show Changing careers

templates/applicantForm.phtml
```diff
@@ -164,6 +183,12 @@
 
                 <div id="onboarding" class="collapse">
 
+                    <label for="fee">Course Fee</label>
+                    <div class="input-group">
+                        <span class="input-group-addon">&pound;</span>
+                        <input id="fee" name="fee" class="submitApplicant form-control" type="number" placeholder="0" min="0" max="10000" step="1" value="<?php echo $applicant->getFee();?>">
+                    </div>
+
```
ashleycoles, line 191: I seem to have an issue here where I'm getting the "Total payment is more than course price" error when doing this:Screenshot 2021-05-26 at 15 14 50

ashleycoles (dismissed):

src/Validators/ApplicantValidator.php
```diff
@@ -44,7 +44,7 @@ public static function validateAdditionalFields(array $applicant): bool
         if (
             (
                 (int)$applicant['upfront'] + (int)$applicant['edaid'] + (int)$applicant['diversitech']
-            ) > Globals::ACADEMYPRICE
+            ) > (int)$applicant['fee']
```
mporam, line 47: what if there is no applicant fee? its an optional field so if its blank, this validation should use the academyprice global instead

mporam (changes requested):

mporam (approved):

PR #352, opened 2021-05-27
Story 4 - Fix FE Applicant form validation messages - stop back end "invalid data" messages leaking through

imypullan:
FE & BE validation re applicant adding/editing cross-checked and fixed to be consistent.
BE validation messages changed to be more valuable (although they should no longer make it through to the FE).
FE validation messages changed to be more descriptive of what the user needs to do to satisfy validation requirements.
Tests added for new validator.


public/js/validations.js
```diff
@@ -1,5 +1,5 @@
 function isDate(date) {
-    let pattern = /([12]\d{3})-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])/;
+    let pattern = /([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/;
```
mporam, line 2: why?

src/Validators/NameValidator.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Portal\Validators;
+
+use Exception;
+
+class NameValidator
```
mporam, line 7: can we rename this? it could be used to validate things that are not names, make it more generic as alpha validator perhaps? it could also go into the string validator class probably?

mporam (changes requested):

public/js/validations.js
imypullan: This makes it exactly the same as the BE vallidation

src/Validators/StringValidator.php
```diff
@@ -6,6 +6,7 @@ class StringValidator
 {
     public const MAXVARCHARLENGTH = 255;
     public const MAXTEXTLENGTH = 10000;
+    private const PATTERN = "/^[a-z ,.'-]+$/i";
```
mporam, line 9: not a great name for a constant given the generic nature of this validator

src/Validators/StringValidator.php
```diff
@@ -50,4 +51,20 @@ public static function validateLength(
         }
         return true;
     }
+
+    /**
+     * Make sure the Name is valid
```
mporam, line 56: comment is wrong

mporam (changes requested):

src/Validators/StringValidator.php
imypullan: fixed

src/Validators/StringValidator.php
imypullan: fixed

mporam (dismissed):

mporam (approved):

PR #374, opened 2021-07-28
Assessment day options changed into a dropdown list

AlicjaZub:


public/js/populateDropdowns.js
```diff
@@ -100,6 +100,31 @@ const checkedCohortDates = () => {
...
+        let dateOptions = {year: 'numeric', month: 'long', day: 'numeric'};
+        assessmentOptions += `value="${
+            item.id
+        }">${date.toLocaleDateString(
+            'en-GB',
+            dateOptions
+        )}</option>`;
+      });
+
+      element.innerHTML += assessmentOptions;
```
mporam, line 123: there is no unselected assessment option. What if we want to edit an applicant and not set an assessment date yet?

mporam (changes requested):

mporam (approved):

PR #378, opened 2021-07-29
Updated the dates cohorts dates shown, added an icon

AlicjaZub:


src/Models/ApplicantModel.php
```diff
@@ -136,11 +139,19 @@ public function getApplicants(
...
         $query->bindValue(':numberPerPage', $this->numberPerPage, \PDO::PARAM_INT);
         $query->execute();
         $applicants = $query->fetchAll();
-        foreach ($applicants as $applicant) {
+        foreach ($applicants as $key => $applicant) {
+            if (
+                !empty($applicant->getChosenCourseId()) && $cohortId !== '%' &&
+                $applicant->getChosenCourseId() !== $cohortId
+            ) {
+                unset($applicants[$key]);
```
mporam, line 152: can you add a comment above this line that explains what this is doing. this code is not very clear so needs an explanation why this is being done

mporam (changes requested):

mporam (approved):

PR #379, opened 2021-09-27
added reload function to next stage function

Freschwamm:
Added location.reload(); to next stage function. Line 79. Only code that has been changed or added.


SamBellDev (approved):
Nice simple fix!

PR #383, opened 2021-09-29
S3 t2 show applicant data in pop up

ros-a:
Added option for custom assessment day to the applicant form, and added customAssessmentDay field in database. Also made sure the chosen custom assessment day appears in the pop-up of each applicant.


SamBellDev (approved):
Works for me in Firefox and Chrome.

PR #392, opened 2021-11-23
S3t1 added gender fields

Roomsycakes01:


src/Controllers/FrontEnd/StudentApplicationFormPageController.php
```diff
@@ -34,8 +34,12 @@ public function __construct(ApplicationFormModel $model, PhpRenderer $view)
     public function __invoke(RequestInterface $request, ResponseInterface $response, array $args): ResponseInterface
     {
         $data = [
-            'backgroundInfo' => $this->model->getBackgroundInfo(),
-            'hearAbout' => $this->model->getHearAbout(),
+            'data' => [
+                'backgroundInfo' => $this->model->getBackgroundInfo(),
+                'hearAbout' => $this->model->getHearAbout(),
+                'Cohorts' => $this->model->getCohorts(),
```
jesshamby, line 40: Not sure if this is because of how it is written elsewhere but think 'cohorts' and 'gender' would be more consistent casing

jesshamby (approved):
Looks spot on - just a single casing comment :)

PR #399, opened 2021-11-24
Story9 stage flags

matthewsearle01:


src/Validators/StageValidator.php
```diff
@@ -29,7 +34,13 @@ public static function validateExistingStage(array $stage)
             is_numeric($stage['id']) &&
             !empty($stage['title']) &&
             !empty($stage['order']) &&
-            is_bool($stage['student'])
+            (is_bool($stage['student']) ||
```
nealioa, line 37: The formatting of this is not consistent with that in validateNewStage().

nealioa (changes requested):

nealioa (approved):

PR #404, opened 2021-11-24
Story13 prevent deleting stages with assignees

Judylj:


src/Entities/CompleteApplicantEntity.php
```diff
@@ -159,9 +159,7 @@ public function getAssessmentDay(): ?string
         return $this->assessmentDay;
     }
 
-    /**
```
nealioa, line 162: Why was this deleted? This getter is now inconsistent with the others in this file.
Also there are now two blank lines between getters - only one is needed.

src/Entities/OptionsEntity.php
```diff
@@ -62,4 +63,13 @@ public function getOptionDeleted(): int
     {
         return $this->deleted;
     }
+
+    /**
+     * Returns number of assignees to option
```
nealioa, line 68: Add a blank between these two like the other docs above.

src/Entities/StageEntity.php
```diff
@@ -94,4 +95,13 @@ public function getStudent(): ?bool
     {
         return $this->student;
     }
+
+    /**
+     * Returns number of assignees to stage
```
nealioa, line 100: Add a blank between these two like the other docs above.

nealioa (changes requested):

nealioa (approved):

PR #405, opened 2021-11-25
S11t2

jesshamby:
Please take a look at the changes to eventsPage.js. 
N.B. s10 has been merged in so there is extra stuff the other team have written so pls only look at lines 213-230 as these are the lines for this task :)


Roomsycakes01 (commented):
looks good!You just need to get it past the linter now :)

PR #407, opened 2021-11-25
S15t2 lock event listener

mclack:
Adds event listener for stageLocks to disable stage delete buttons when locking and triggers allowDeletionModal when unlocking


DevUsman-18:
looks good

PR #413, opened 2021-11-25
s11t3

jesshamby:
Made general styling imporvements, changed colors to iO banding, increased size of event name relative to the date and time in the header and made the labels in the more info dropdown div bold


Mondo15 (approved):
😀😀

PR #415, opened 2021-11-25
S11

jesshamby:


src/Models/ApplicantModel.php
```diff
@@ -521,4 +521,15 @@ public function updateApplicantStageAndOptionIds(int $applicantId, int $stageId,
...
+    public function getAssessmentApplicants(): array
+    {
+        $stmt = "SELECT `applicants`.`id`, `applicants`.`name`, `applicants`.`email`, 
+                    `applicants_additional`.`assessmentDay` 
+                FROM `applicants` INNER JOIN `applicants_additional` 
+                ON `applicants`.`id`=`applicants_additional`.`id`;";
+        $query = $this->db->prepare($stmt);
+        $query->execute();
+        return $query->fetchAll();
+    }
```
ashleycoles, line 534: This query doesn't account for deleted applicants.

public/js/eventsPage.js
```diff
@@ -208,24 +210,55 @@ async function displayHiringPartnersAttending(event){
...
  * @param events an object which contains information about an event
  */
-async function eventGenerator(event, hiringPartners) {
+async function eventGenerator(event, hiringPartners, applicants) {
     let eventInformation = ''
     let date = new Date(event.date).toDateString()
+    let numberOfAttendees
+    if (event.category_name === 'Assessment') {
+        numberOfAttendees = applicants.length
+        console.log(numberOfAttendees)
```
ashleycoles, line 219: This is actually counting all applicants, not just ones assigned to a given assessment day, is this because it's from story 10 which was unfinished?

public/js/eventsPage.js
```diff
@@ -208,24 +210,55 @@ async function displayHiringPartnersAttending(event){
...
+    if (event.category_name === 'Assessment') {
+        // put code for assessment event type here!
+        eventInformation += '<div>';
+        eventInformation += '<table class="col-xs-12 table-bordered table">';
+        eventInformation += '<tr>';
+        eventInformation += '<th class="col-xs-2">Name</th>';
+        eventInformation += '<th class="col-xs-3">Email</th>';
+        eventInformation += '</tr>';
+            applicants.forEach((applicant) => {
+                if (applicant.assessmentDay == event.id) {
```
ashleycoles, line 253: I think you need a check like this when counting applicants, but again this probs falls under s10?

ashleycoles (changes requested):

public/js/eventsPage.js
jesshamby: Discussed but have added this in the new commit as the S10 query to the DB doesn't currently only fetch applicants attending a given event

public/js/eventsPage.js
jesshamby: Discussed with you but yes, didn't realise the query in s10 fetches all applicants and then filters them in the js file instead of fetching applicants for a specific event

public/js/eventsPage.js
```diff
@@ -208,24 +210,59 @@ async function displayHiringPartnersAttending(event){
...
+async function eventGenerator(event, hiringPartners, applicants) {
     let eventInformation = ''
     let date = new Date(event.date).toDateString()
+    let numberOfAttendees
+    if (event.category_name === 'Assessment') {
+        let assessmentApplicants = applicants.filter(function (applicant) {
+            if (applicant.assessmentDay == event.id) {
+                return applicant
+            }
+            })
```
ashleycoles, line 222: This works better. Because filter only needs to return a boolean value (if true, it keeps the item, if false it removes it) you could simply the logic here like this:
```
let assessmentApplicants = applicants.filter(function (applicant) {
    return applicant.assessmentDay === event.id
})
```

If you used a fat arrow function you could make it even more compact:

```
let assessmentApplicants = applicants.filter(applicant => applicant.assessmentDay === event.id)
```

ashleycoles (changes requested):

public/js/eventsPage.js
jesshamby: That is much prettier! JS seems like so long ago now :'(

src/Models/ApplicantModel.php
jesshamby: Will speak to the people on s10 who wrote this!

ashleycoles (approved):

PR #417, opened 2021-11-25
S3t3

Roomsycakes01:


public/js/studentApplicationForm.js
```diff
@@ -13,16 +15,168 @@ nextButtons.forEach(($prevButton) => {
...
+            output = pageOneValidation(formWrapper)
+            break;
+        default:
+            output = true
+    }
+    return output
+}
+
+function pageOneValidation(formWrapper) {
+    let inputs = formWrapper.querySelectorAll('input')
```
jesshamby, line 79: could these first 6 lines be combined into a separate function and called in each page validator to make the code more DRY?

jesshamby (approved):
Some v v nice code - made one comment which might help make the validation code a bit more DRY :)

PR #426, opened 2022-01-20
Converting to SCSS

ashleycoles:
Adding basic SCSS with some variables and nesting as well as tidying up templates so they no longer need to manually import a bunch of CSS files.


README.md
```diff
@@ -13,6 +13,9 @@ Live Version at: https://portal.dev.io-academy.uk/
 2. Create database with name `academyPortal` and populate using latest version in db/
 3. To run the application locally: `composer start`
 
+### Compiling SCSS
+- Run `composer watch` in the root of the project
```
charliecog, line 17: can we add 'must have sass installed globally' before what is written here please

charliecog (commented):
as mentioned in the thread, if its possible to strip out generated css and throw in a gitignore, might make it a little more predictable for the students.
apart from that its good to go!

charliecog (approved):
looks good to go!

PR #428, opened 2022-04-04
Enabled a modul to open on the events page when an applicant's name is clicked

maxwell-01:
contained jQuery for applicant modal in a function called display app…licant modal, updated script tags to support import export, imported applicant modal to eventspage.js

### DIFF
--- a/public/js/applicantModal.js
+++ b/public/js/applicantModal.js
@@ -180,7 +180,3 @@ export function displayApplicantModal() {
 })
 }
 displayApplicantModal()
-
-
-
-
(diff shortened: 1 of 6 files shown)


public/js/applicantModal.js
```diff
@@ -178,3 +178,9 @@ $(document).ready(function () {
         $("#applicantModal").modal()
     })
 })
+}
+displayApplicantModal()
+
+
+
+
```
RichVelo, line 186: whitespace needed?

PR #434, opened 2022-04-06
S7 applicants pagination does not work

maxwell-01:
bug fixed, pagination works

### DIFF
--- a/src/Controllers/FrontEnd/ApplicantsPageController.php
+++ b/src/Controllers/FrontEnd/ApplicantsPageController.php
@@ -14,6 +14,7 @@ class ApplicantsPageController extends Controller
     private $renderer;
     private $applicantModel;
     private $stageModel;
+    private int $numberOfApplicantsPerPage = 20;
 
     /**
      * ApplicantsPageController constructor.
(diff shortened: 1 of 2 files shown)


src/Models/ApplicantModel.php
```diff
@@ -137,25 +121,21 @@ public function getApplicants(
...
         $query->bindValue(':name', $name);
         $query->bindValue(':stageId', $stageId);
         $query->bindValue(':cohortId', $cohortId);
         $query->bindValue(':cohortId2', $cohortId);
-        $query->bindValue(':offsets', $offset, \PDO::PARAM_INT);
-        $query->bindValue(':numberPerPage', $this->numberPerPage, \PDO::PARAM_INT);
         $query->execute();
-        $applicants = $query->fetchAll();
-        foreach ($applicants as $key => $applicant) {
+        $allApplicants = $query->fetchAll();
```
norayna, line 131: no need to rename this from `$applicants`

src/Controllers/FrontEnd/ApplicantsPageController.php
```diff
@@ -14,6 +14,7 @@ class ApplicantsPageController extends Controller
     private $renderer;
     private $applicantModel;
     private $stageModel;
+    private int $numberOfApplicantsPerPage = 20;
```
norayna, line 17: unfortunately you have to delete that type hint 😭

src/Models/ApplicantModel.php
maxwell-01: changed back

norayna (changes requested):
it needs to redirect when you go to a page number that doesn't exist e.g. `?page=8` or `?page=-8`

PR #443, opened 2022-05-30
S2t1 styling

Mead-y:


doubt-ur-vibe (commented):
Some of the styling is breaking at smaller sizes, specifically the radio buttons and question mark pop-up. The media queries are a bit inconsistent, you might want to consider using a container class with both max-width and width properties to make the responsiveness more consistent. Great work though, the form is definitely looking better :)

PR #449, opened 2022-05-30
Story4 filter events

haefnir:


src/Models/EventModel.php
```diff
@@ -169,24 +169,77 @@ public function searchPastEvents(string $searchTerm): array
...
+        return $query->fetchAll();
+    }
+
+    /**
+     * Gets past events based on an optional categoryId and searchQuery
+     * @param string|null $categoryId
+     * @param string|null $searchQuery
+     * @return array
+     */
+    public function getPastEventsByCategoryIdAndSearch(string $categoryId = null, ?string $searchQuery = ''): array
```
ashleycoles, line 221: $categoryId should be nullable as well - does $searchQuery need to be nullable?

ashleycoles (changes requested):

ashleycoles (approved):

PR #457, opened 2022-05-31
S3t2 local storage

haefnir:


public/js/saveApplicant.js
```diff
@@ -40,6 +41,11 @@ document.querySelector('#submitApplicant').addEventListener('click', e => {
     }
 });
 
+document.querySelector('#saveProgress').addEventListener('click', e => {
```
ashleycoles, line 44: There are actually multiple saveProgress buttons, this only works for the first one. What you'll need to do is change it so the save buttons have a class instead, then use a querySelectorAll to get all of them

public/js/saveApplicant.js
```diff
@@ -73,6 +79,91 @@ let errorMessage = (validationType) => {
...
+                    stagedData[formItem.name] = formItem.checked;
+                }
+            } else {
+                stagedData[formItem.name] = formItem.value;
+            }
+        }
+    })
+    Object.keys(stagedData).forEach(stagedItemKey => {
+        console.log(stagedItemKey, stagedData[stagedItemKey])
+        window.localStorage.setItem(stagedItemKey, stagedData[stagedItemKey])
```
ashleycoles, line 103: No need to do window.localstorage, the window object is the global object which means you can just do `localstorage.setItem()`

ashleycoles (changes requested):

PR #462, opened 2022-08-09
Autowiring

ashleycoles:


mporam (commented):
code looks fine in general, although with the removal of the interface typehints for some of the classes, is there any benefit in having the interfaces anymore? Are they being used? if not, should we delete them?

PR #469, opened 2022-09-27
s1t4cut-trainers-page-controller-functionality

mikeycodingstuff:
Updated controller functionality and factory to account for trainer model functionality. Added trainer model to DIC


aidanmaskell (approved):
looks great!

PR #471, opened 2022-09-27
Surf s3

Harry-Johnston:


public/js/populateDropdowns.js
```diff
@@ -204,3 +206,24 @@ let outputBackgroundInfo = (options) => {
...
 };
+
+let checkPastAssessmentCheckbox = (futureDates, allDates) => {
+  pastAssessmentCheckbox.addEventListener('change', e => {
+    if (e.target.checked === true) {
+      outputAssessmentDates(allDates)
+    }
+    if (e.target.checked === false) {
+      outputAssessmentDates(futureDates)
+    }
```
ashleycoles, line 217: Make this into an else

src/Models/EventModel.php
```diff
@@ -169,7 +169,7 @@ public function searchPastEvents(?string $searchTerm): array
      * @return array An array of Events based on category ID and previous X months from current date.
      */
 
-    public function getEventsByCategoryId(string $categoryId, int $previousMonths = 0): array
+    public function getEventsByCategoryId(string $categoryId, int $previousMonths = 9999): array
```
ashleycoles, line 172: I don't like this. Is there a better way?

ashleycoles (changes requested):
Looking pretty good.

Seems to me like you've missed this requirement: 

If an applicant is assigned to a past date when editing them. The date should display as a uneditable text above the drop down

src/Models/EventModel.php
ashleycoles: Would suggest leaving the original default, but pass in the number of previous months you want when this method gets called in the controller.

Maybe don't go for 9999, maybe do like 5 years worth?

templates/applicantForm.phtml
```diff
@@ -133,9 +133,12 @@
         </div>
 
         <div id="assessment" class="collapse">
-            <label for="assessmentDay">Assessment day</label>
+            <label for="assessmentDay">Assessment day
+                <label for="pastAssessmentDays">
+                    <input type="checkbox" id="pastAssessmentDays" name="accept" value="yes">  Include past assessment days
```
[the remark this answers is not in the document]
Harry-Johnston: done

public/js/populateDropdowns.js
Harry-Johnston: done

src/Models/EventModel.php
Harry-Johnston: done

templates/applicantForm.phtml
```diff
@@ -134,8 +134,13 @@
 
         <div id="assessment" class="collapse">
             <label for="assessmentDay">Assessment day</label>
+            <p>
+                <label for="pastAssessmentDays">
+                    <input type="checkbox" id="pastAssessmentDays" name="accept" value="yes">  Include past assessment days
+                </label>
+            </p>
+            <p>Current saved date: <?php echo $applicant->getAssessmentDay();?></p>
```
ashleycoles, line 142: Please could you only display this if there is a date currently set? Also, wondering if it's possible only to display it if the date is a past one?

ashleycoles (changes requested):
One last request!

ashleycoles (approved):

ashleycoles (approved):

PR #501, opened 2022-11-24
S1t2 show error alerts below input fields

O-MAY:


public/js/addHiringPartnerContact.js
```diff
@@ -1,20 +1,39 @@
...
+                message.classList.add('hidden')
+                break;
+            }
+        }
+    });
 
-    if(validateAddContactForm(data)){
-        fetch('./api/addContact', {
+    if (formIsValid) {
+        // send it!
```
TherealGuah, line 35: Remove comment from final build please!

TherealGuah (changes requested):

PR #507, opened 2023-04-03
s3t1-save url bug fixed

Steph-ski:
- identify bug and fix


alreuben (approved):

Yas-Shen (commented):
Looks good to me too.

PR #511, opened 2023-04-04
S5t2 create event listener

nesclark:


Yas-Shen (commented):
Looks good, regEx was perfect when tested thoroughly.

PR #516, opened 2023-04-04
S2t8 editable student profile update db

melody-sylvestre:
Back End - Create method that will update the database: Models/ApplicantModel::updateStudentProfileFields()


west-country (approved):
all good

PR #517, opened 2023-04-04
S2 editable student profile t7 sanitise student profile fields

DanAbsolon:
started test stuff but ignore that please. Please just review the stuff in sanitiseApplicant


west-country (approved):
Looks good. 

Looks like you can also use (int) to convert boolean into 1 or 0 at line 134, but only mention it because I was looking up (int).

No need for changes.

west-country (approved):
all good

PR #520, opened 2023-04-05
S5t3 sanitisation and validation php

Joperes-ym:
CourseValidator - validate length and that its a number
Review unit tests and update as necessary


tests/Validators/IntegerValidatorTest.php
```diff
@@ -0,0 +1,31 @@
...
+        $input = 10009;
+        $this->expectException(\Exception::class);
+        IntegerValidator::validateInteger($input, 1, 999);
+    }
+
+    public function testMalformedIntegerValidator()
+    {
+        $input = "Helloooo";
+        $this->expectException(\TypeError::class);
+        IntegerValidator::validateInteger($input, 1, 999);
```
HadertS, line 29: It would be good to include `$this->getExpectedExceptionMessage('Number out of range');` for a more accurate representation of its actual use.

tests/Validators/IntegerValidatorTest.php
```diff
@@ -0,0 +1,31 @@
...
+        $expected = 5;
+        $input = 5;
+        $case = IntegerValidator::validateInteger($input, 1, 999);
+        $this->assertEquals($expected, $case);
+    }
+
+    public function testFailureIntegerValidator()
+    {
+        $input = 10009;
+        $this->expectException(\Exception::class);
```
HadertS, line 21: It would be good to include $this->getExpectedExceptionMessage('Number out of range'); for a more accurate representation of its actual use.

tests/Validators/IntegerValidatorTest.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Tests\Validators;
+
+use Portal\Validators\IntegerValidator;
+use Tests\TestCase;
+
+class IntegerValidatorTest extends TestCase
```
HadertS, line 8: Add a docblock

tests/Validators/IntegerValidatorTest.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function testSuccessIntegerValidator()
```
HadertS, line 10: Docblock

tests/Validators/IntegerValidatorTest.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $expected = 5;
+        $input = 5;
+        $case = IntegerValidator::validateInteger($input, 1, 999);
+        $this->assertEquals($expected, $case);
+    }
+
+    public function testFailureIntegerValidator()
```
HadertS, line 18: Docblock

src/Validators/IntegerValidator.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * Validates that an integer is within the $min and $max range
+     *
+     * @throws \Exception if number out of range
+     */
+    public static function validateInteger(int $unvalidated, int $min, int $max): int
+    {
+        $options = ["options" => ["min_range" => $min, "max_range" => $max]];
+        if (filter_var($unvalidated, FILTER_VALIDATE_INT, $options) !== false) {
```
Steph-ski, line 15: !== false
is this a double negative?!

src/Validators/CourseValidator.php
Steph-ski, line 1: do we need to add in a doc block for the validate() function? Aware we didn't write most of it, just amended functionality.

src/Validators/IntegerValidator.php
HadertS: Thats not equal to false, so its not a double negative.

HadertS (changes requested):
Some changes, steph has picked up a lot of things I also noted. There is also an issue with the variable naming in AddCourse.js but this can be fixed as part of the story 5 merge.

tests/Validators/IntegerValidatorTest.php
Yas-Shen: It doesn't work, mate.

HadertS (approved):

PR #522, opened 2023-04-05
S2 editable student profile

robin-hodge:


templates/studentProfile.phtml
```diff
@@ -5,10 +5,12 @@
     require 'partials/layout/head.phtml';
     echo \Portal\ViewHelpers\LayoutViewHelper::outputTitle('Applicants');
     ?>
+    <script>const studentId = <?=intval($id);?></script>
```
norayna, line 8: why is this not a line inside the file updateStudentProfile.js?

public/js/updateStudentProfile.js
```diff
@@ -0,0 +1,69 @@
...
+        containerDiv.innerHTML = 
+        '<form class="form studentProfileEditableField">' +
+        '<label for="' + fieldName + 'TextBox">' + descriptionTag.textContent + '</label>' +
+        '<input type="text" id="' + fieldName + 'TextBox" name="' + fieldName + '">' + 
+        '<input class="saveButton  btn btn-primary btn-sm" type="submit" value="Save">' +
+        '</form>'
+    }
+
+    const saveButton = document.querySelector('.saveButton')
+    const form = document.querySelector('.form')
```
norayna, line 36: so you're using a class called just "form"? i think you need a more descriptive name

public/js/updateStudentProfile.js
```diff
@@ -0,0 +1,69 @@
...
+        } 
+        const jsonUpdatedField = JSON.stringify(updatedField)
+        fetch('/api/updateStudentProfile', {
+            method: 'PUT',
+            body: jsonUpdatedField,
+            headers: {
+                'Content-Type': 'application/json'
+            }
+        }).then( async (response) => {
+            return response.json().then((data) => {
```
norayna, line 54: as far as i can tell, you don't need to return here and you also don't need to make the anon arrow function async

norayna: actually i think this whole block is a bit weird 🤔 i'll figure out shortly what i want you to do with it (probably just make it standard as per the notes)

public/js/updateStudentProfile.js
```diff
@@ -0,0 +1,69 @@
+const edaidEditButton = document.querySelector('#edaidEditButton')
+const upfrontEditButton = document.querySelector('#upfrontEditButton')
+const laptopEditButton = document.querySelector('#laptopEditButton')
+const githubUsernameEditButton = document.querySelector('#githubUsernameEditButton')
+
+function handleEditClick(event) {
```
norayna, line 6: this is one massive function, needs to be split into smaller ones

public/js/updateStudentProfile.js
tauroid: agree, I suggested it just to get it working but imo await is the way to go

src/Models/ApplicantModel.php
```diff
@@ -558,4 +558,38 @@ public function appendToAssessmentNotes(int $id, string $assessmentNote): bool
...
+
+    public function getFeePaymentMethods(array $userInput): array
+    {
+        $query = $this->db->prepare('SELECT diversitech, edaid, fee, upfront FROM applicants_additional
+                             WHERE id=?');
+        $query->execute([$userInput['id']]);
+        $feePaymentMethod = $query->fetch();
+        return  $feePaymentMethod;
+    }
 }
```
norayna, line 595: for all these methods, it would be better practice to list out separate parameters, rather than having one parameter called userInput. this is because:
- this relies on another part of the system providing the info in the correct format - essentially "tight coupling"
- it makes the code more opaque (harder to tell what the method actually needs to use)
you can see other methods in this model class don't do this, other than the ones that take $applicant which has a million fields

src/Models/ApplicantModel.php
```diff
@@ -558,4 +558,38 @@ public function appendToAssessmentNotes(int $id, string $assessmentNote): bool
...
+    public function updateGithubUsername(array $userInput): bool
+    {
+        $query = $this->db->prepare('UPDATE `applicants_additional` SET githubUsername=:githubUsername
+         WHERE id=:id');
+        return $query->execute($userInput);
+    }
+
+    public function updateLaptop(array $userInput): bool
+    {
+        #NB:laptop should be 0 or 1 apparently
```
norayna, line 576: what am i to do with this information? 😅

src/Controllers/API/UpdateStudentProfileController.php
```diff
@@ -0,0 +1,182 @@
...
+    public function __invoke(Request $request, Response $response, array $args): Response
+    {
+        $updatedStudentProfileData = $request->getParsedBody();
+
+        if (
+            !empty($_SESSION['studentLogin']) &&
+            $_SESSION['studentLogin'] &&
+            $_SESSION['studentId'] == $updatedStudentProfileData['id'] ||
+            !empty($_SESSION['loggedIn']) &&
+            $_SESSION['loggedIn']
```
norayna, line 32: it's not entirely clear here which operator takes precedence because you have one || among lots of &&s, so it would help to put some brackets to make it clear which conditions go together

src/Controllers/API/UpdateStudentProfileController.php
```diff
@@ -0,0 +1,182 @@
...
+        ) {
+            $responseBody = [];
+
+            if (array_key_exists("edaid", $updatedStudentProfileData)) {
```
norayna, line 36: i would like the remainder of the method __invoke to look something like this:
```php
    if (array_key_exists("edaid", $updatedStudentProfileData)) {
        $this->updateEdaid($updatedStudentProfileData);
    } elseif (array_key_exists("upfront", $updatedStudentProfileData)) {
        $this->updateUpfront($updatedStudentProfileData);
    } elseif (array_key_exists("githubUsername", $updatedStudentProfileData)) {
        $this->updateGithubUsername($updatedStudentProfileData);
    } elseif (array_key_exists("laptop", $updatedStudentProfileData)) {
         $this->updateLaptop($updatedStudentProfileData);
    } else {
        $responseBody["success"] = false;
        $responseBody["msg"] = "Unsupported field";
        $responseBody["data"] = [];
        $statusCode = 400;
        return $this->respondWithJson($response, $responseBody, $statusCode);
    }
}
return $response->withStatus(401);
```

basically i want you to break up your huge method into multiple smaller methods so the code is more readable

templates/studentProfile.phtml
norayna, line 8: it would be better to store this as a data attribute somewhere on the page and then retrieve it in the js using dataset

src/Validators/ApplicantValidator.php
```diff
@@ -66,15 +66,7 @@ public static function validateAdditionalFields(array $applicant): bool
...
-        $feePaymentMethods = (int)$applicant['upfront'] + (int)$applicant['edaid'] + (int)$applicant['diversitech'];
-
-        if (
-            ((int)$applicant['fee'] > 0
-            && $feePaymentMethods > (int)$applicant['fee'])
-            || ($feePaymentMethods > Globals::ACADEMYPRICE)
-        ) {
-            throw new \Exception('Total payment is more than course price');
-        }
+        ApplicantValidator::validateFeePaymentMethods($applicant);
```
norayna, line 69: this whole file is chef's kiss

src/Sanitisers/ApplicantSanitiser.php
```diff
@@ -102,4 +102,48 @@ public static function sanitiseAdditionalFields(array $applicant): array
...
     }
+
+    /**
+     * Sanitise edaid for editing student profile
+     *
+     * @return ?int sanitised $savedProfileFieldEdAid
+     */
+    public static function sanitiseEdAid($savedProfileFieldEdAid): ?int
+    {
+        return $savedProfileFieldEdAid ? (int)$savedProfileFieldEdAid : null;
```
norayna, line 113: this for all your functions:
it's weird that you're checking whether the field is there inside the sanitising function. it would be better to only attempt to sanitise the field if it's there. otherwise why would you pass an empty value to a function, then figure out that it's empty and decide to do nothing? you could have checked before calling the function.

norayna: as a result of that comment, all that would remain in this function is a typecast to int: (int) $savedProfileFieldEdAid. so then it is questionable whether this function should exist at all

nealioa (approved):

PR #523, opened 2023-04-05
S5 add available places to courses

randallcoding:
As a: academy admin
I want: keep track of how many places are available on each course
So that: I know how many places I can assign applicants to

‌

Requirements:

When adding a new course, provide a field for specifying how many places are available
Must be able to specify how many remote and in-person places
Must not be 0 or negative
Related field must not be available if the course format does not match. i.e dont show the amount of remote spaces field if the course is not available remote


public/js/addCourse.js
```diff
@@ -1,5 +1,26 @@
...
+        in_person_spaces.classList.remove('hidden');
+    } else {
+        in_person_spaces.classList.add('hidden');
+    }
+
+})
+
+courseForm.addEventListener('change', () => {
+    if (remote_checkbox.checked) {
+        remote_spaces.classList.remove('hidden');
```
mporam, line 18: theres a better way to do this, look up `classList.toggle`

src/Models/CourseModel.php
```diff
@@ -46,29 +46,37 @@ public function addCourse(array $newCourse): string
...
+            :remote_spaces);");
 
         $startDate = $newCourse['startDate'];
         $endDate = $newCourse['endDate'];
         $name = $newCourse['courseName'];
         $notes = $newCourse['notes'];
         $in_person = $newCourse['in_person'];
         $remote = $newCourse['remote'];
+        $in_person_spaces = $newCourse['in_person_spaces'];
+        $remote_spaces = $newCourse['remote_spaces'];
```
mporam, line 70: these variables are totally pointless, just use the array with the correct assoc key in the bindParam calls and remove these

templates/addCourse.phtml
```diff
@@ -39,14 +39,20 @@ $trainersCheckbox = \Portal\ViewHelpers\TrainerCheckboxViewHelper::displayTraine
         <label for="course-notes">Notes</label>
         <textarea name="notes" class="create-courses form-control" placeholder="Notes" data-max="5000" maxlength="5000"></textarea>
         <div id="courseNotesError" class="alert hidden formItem_alert"></div>
+
         <div class="form-group">
             <label for="in_person">In Person?</label>
             <input type="checkbox" id="in_person" value="1" class="create-courses" name="in_person">
+            <input id="in_person_spaces" name="in_person_spaces" class="create-courses form-control hidden" type="text" data-required="true" />
```
mporam, line 46: why `type="text"` and not `type="number"` ?

templates/addCourse.phtml
```diff
@@ -39,14 +39,20 @@ $trainersCheckbox = \Portal\ViewHelpers\TrainerCheckboxViewHelper::displayTraine
...
+            <div id="in_person_spacesError" class="alert hidden formItem_alert"></div>
         </div>
         <div class="">
             <label for="remote">Remote?</label>
             <input type="checkbox" id="remote" value="1" class="create-courses" name="remote">
+            <input id="remote_spaces" name="remote_spaces" class="create-courses form-control hidden" type="text" data-required="true" />
```
mporam, line 52: same question

mporam (changes requested):

templates/addCourse.phtml
mporam, line 46: also this is missing a label? am I just supposed to psychically know what to use this input box for?

db/academyPortal.sql
```diff
@@ -26,50 +26,49 @@
...
+  `name` varchar(255) DEFAULT NULL,
+  `email` varchar(255) DEFAULT NULL,
+  `phoneNumber` varchar(20) DEFAULT NULL,
+  `whyDev` text,
+  `codeExperience` text,
+  `hearAboutId` int(11) DEFAULT NULL,
+  `eligible` enum('1','0') DEFAULT NULL,
+  `eighteenPlus` enum('1','0') DEFAULT NULL,
+  `finance` enum('1','0') DEFAULT NULL,
+  `notes` text,
```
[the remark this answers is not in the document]
HadertS: https://dev.mysql.com/doc/refman/8.0/en/blob.html
BLOB and TEXT columns cannot have DEFAULT values.

mporam (approved):

PR #530, opened 2023-09-25
S1t2 fix backend error message not displayed caterpillars

CMeier556:
Error messages displays when back end fails (changing the db name is one way of testing this). Fixed by removing a hidden attribute that was being given to the message upon the form not being processed by the server.

Other way to test this is to turn off front-end validation and check console that back-end validation errors still work.


alexjames4 (approved):
appears to be fixed :)

PR #532, opened 2023-09-25
S3 t1 add lock checkbox applicant edit page caterpillars

CosminDarius1:
the lock checkbox changes from locked to unlock and vice versa

doesn’t actually change editability of the fields from the students perspective

### DIFF
--- a/public/js/lockToggle.js
+++ b/public/js/lockToggle.js
@@ -4,4 +4,5 @@
          toggleLock.addEventListener('click',(e) => {
              e.target.classList.toggle("bi-lock");
              e.target.classList.toggle("bi-unlock");
-         })});
+         })
+     });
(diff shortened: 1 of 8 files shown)


public/js/lockToggle.js
```diff
@@ -0,0 +1,7 @@
+
+ const toggleLocks = document.querySelectorAll('#lockToggle')
+     toggleLocks.forEach(toggleLock => {
+         toggleLock.addEventListener('click',(e) => {
+             e.target.classList.toggle("bi-lock");
+             e.target.classList.toggle("bi-unlock");
+         })});
```
alexjames4, line 7: i think the second set of }); needs to be on line 8 in line with the toggleLocks.forEach on line 3

alexjames4 (changes requested):
As discussed in person move padlock from laptop deposit paid to Github Username

PR #535, opened 2023-09-26
S5t2 fix profile link bug

CMeier556:
- Bug was fixed by changing the studentUrl variable in applicantModal.js on line 56 to the profile page link
- Then in order for it to read the student's id for the proper link, had to concatenate data.id on the end of the link (line 69)
- Added an if statement (is not currently necessary) so if there is no link to an applicant's profile it will tell you "No Link Yet"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@ sendgrid.env
 .env
 .circleci/.config.yml.swp
 .phpunit.result.cache
+.phpunit.cache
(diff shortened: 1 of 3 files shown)


alexjames4:
How do I replicate bug? please tell me in the pull request summary

alexjames4 (approved):
Bug appears to be fixed, well done

PR #537, opened 2023-09-27
Story5 fix applicant modal profile link bug

alexjames4:
Story Requirements:

"the applicant modal is supposed to display a link to the applicants profile page, but it doesnt work. please fix"

From Colin: 
_Bug was fixed by changing the studentUrl variable in applicantModal.js on line 56 to the profile page link
Then in order for it to read the student's id for the proper link, had to concatenate data.id on the end of the link (line 69)
Added an if statement (is not currently necessary) so if there is no link to an applicant's profile it will tell you "No Link Yet"_

Cheers

### DIFF
--- a/public/js/applicantModal.js
+++ b/public/js/applicantModal.js
@@ -67,7 +67,7 @@ export function addEventListenersToDisplayApplicantModal() {
                         response.json().then(function (data) {
 
                             document.querySelectorAll('#applicantModal section.student').forEach(section => {
-                                if (data.isStudentStage === "1") {
+                                if (data.isStudentStage) {
                                     section.classList.remove('hidden')
                                 } else {
                                     section.classList.add('hidden')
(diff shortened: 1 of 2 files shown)


public/js/applicantModal.js
```diff
@@ -170,7 +170,9 @@ export function addEventListenersToDisplayApplicantModal() {
                             displayField(data, 'additionalNotes', 'No notes')
                             data.chosenCourseDate = prettyDate(data.chosenCourseDate)
                             displayField(data, 'chosenCourseDatePretty', 'Not asked yet')
-                            displayField(data,'userProfileLink', 'No Link Yet' )
+                            if (!studentUrl) {
+                                displayField(data, 'userProfileLink', 'No Link Yet')
+                            }
```
mporam, line 175: given this is unnecessary as it stands, id rather remove this so there isnt code in here that does nothing, as it may confuse other devs

mporam (changes requested):

public/js/applicantModal.js
CMeier556: I gort rid of it

mporam (approved):

PR #554, opened 2023-11-21
add redirection status code + end fix undefined array key error

ndrj19:
- added redirection status code where needed
- fixed undefined array key error => added !empty($_SESSION['loggedIn']) to be able to trigger the redirect if the array key doesn't exist


tuckermitcht (approved):
all good

PR #555, opened 2023-11-21
Brainworms s6 fix redirects in firefox

ndrj19:
- added redirection status code where needed
- fixed undefined array key error => added !empty($_SESSION['loggedIn']) to be able to trigger the redirect if the array key doesn't exist


mporam (changes requested):
change the login check to either a strict comparison or no comparison

mporam (approved):

PR #556, opened 2023-11-22
Brainworms s5t2 dropdown logic

Tiago-Varela4:
added future courses to the dropdown list with a default past course if one selected and added the course name to appear with the date


tuckermitcht (commented):
done

PR #557, opened 2023-11-22
Brainworms s5 only assign students to future course dates

claudia-lim:
Story 5 - Only assign students to future courses and if already on a course this is the default value. The name of the course is also displayed alongside the date in the dropdown options.
https://trello.com/c/SzQppPxc/554-5-only-assign-students-to-future-course-dates


public/js/applicantForm.js
```diff
@@ -15,7 +15,14 @@
     checkedCohortDates();
 
     if (document.querySelector('#chosenCourseId')) {
-        outputCohorts(data.cohorts, document.querySelector('#chosenCourseId'));
+        const chosenCourseId = document.querySelector('#chosenCourseId').dataset.selected;
+        const futureCohorts = filterFutureDates(data.cohorts);
+        if (chosenCourseId) {
+            futureCohorts.push(((data.cohorts.filter((cohort) => {
+                return cohort.id == chosenCourseId;
+            }))[0]));
```
mporam, line 23: what happens if they dont have a chosen course? seems like this might error

mporam, line 23: could this go into its own function perhaps? give it a semantic name so that its clear what its doing

public/js/applicantForm.js
```diff
@@ -15,7 +15,14 @@
     checkedCohortDates();
 
     if (document.querySelector('#chosenCourseId')) {
-        outputCohorts(data.cohorts, document.querySelector('#chosenCourseId'));
+        const chosenCourseId = document.querySelector('#chosenCourseId').dataset.selected;
+        const futureCohorts = filterFutureDates(data.cohorts);
```
mporam, line 19: this is doing a filter, could this be combined with the filter below to save looping the data twice?

mporam (changes requested):

mporam:
also noticed:
- if the chosen course is a future course, it appears in the list twice
- if the chosen course is a past date, it appears at the bottom of the list instead of the top

public/js/applicantForm.js
```diff
@@ -16,13 +16,8 @@
 
     if (document.querySelector('#chosenCourseId')) {
         const chosenCourseId = document.querySelector('#chosenCourseId').dataset.selected;
```
mporam, line 18: this variable isnt needed, but it could contain just the element selector in order to reduce DOM selection

mporam (changes requested):

mporam (approved):

PR #565, opened 2023-11-23
Brainworms s8 create calendar

ndrj19:
- added new dropdown element "Calendar" on homepage next to "Events"
- which takes you to a new calendar page using the [FullCalendar.io](https://fullcalendar.io/) plugin 

- added new frontend controller "CalendarPageController" + route '/calendar' 

- added new route '/api/calendar' + API controller GetAllCalendarEvents for calendar event data
- which is fetched in calendarPage.js to retrieve calendar events and then displayed


app/routes.php
```diff
@@ -134,4 +138,5 @@
     $app->post('/api/sendEmail', SendEmailController::class);
     $app->get('/api/lockApplicantField', LockApplicantFieldController::class);
     $app->get('/api/getAptitudeScore', GetApplicantAptitudeController::class);
+    $app->get('/api/calendar', GetAllCalendarEventsController::class);
```
mporam, line 141: not sure about the route url here, its not descriptive enough. look at the other routes, they are all obvious what they do

db/academyPortal.sql
```diff
@@ -505,7 +505,18 @@ VALUES
...
+    (18,'Hiring Event',1,'Bath','2023-10-28','14:00:00','15:00:00','Jam Sandwich',0),
+    (19,'Other',6,'Bath','2023-11-14','10:00:00','12:00:00',NULL,0),
+    (20,'Graduation',5,'Bath','2023-11-15','14:00:00','15:00:00',NULL,0),
+    (21,'Taster Session',3,'Remote','2023-11-20','14:00:00','15:00:00',NULL,0),
+    (22,'Other',6,'Remote','2023-11-21','19:00:00','22:00:00',NULL,0),
+    (23,'Assessment Day',4,'Remote','2023-11-24','14:00:00','15:00:00',NULL,0),
+    (24,'Sprint Review',2,'Bath','2023-12-10','09:00:00','12:30:00',NULL,0),
+    (25,'Graduation',5,'Bath','2023-12-15','15:00:00','18:00:00','Congratulations!!',0),
+    (26,'Taster Session',3,'Bath','2023-12-25','17:00:00','20:00:00','Chilli family noodles',0),
+    (27,'Other',6,'Bath','2023-12-29','09:00:00','17:00:00',NULL,0);
```
mporam, line 519: can we have some more? use mockaroo to generate atleast 1 event every month for the next few years, ideally multiple

public/js/calendarPage.js
```diff
@@ -0,0 +1,45 @@
+document.addEventListener("DOMContentLoaded", function () {
```
mporam, line 1: dont need this, the file is deferred anyway so will only execute after the content as it is

public/js/calendarPage.js
```diff
@@ -0,0 +1,45 @@
...
+    var calendarEl = document.getElementById("calendar");
+    var calendar = new FullCalendar.Calendar(calendarEl, {
```
mporam, line 3: why vars and not `let` or `const`?

public/js/calendarPage.js
```diff
@@ -0,0 +1,45 @@
...
+            minute: "2-digit",
+            meridiem: true,
+        },
+    });
+
+
+    const extractResponseData = (response) => {
+        let data = response.json();
+        return data;
+    };
```
mporam, line 16: this doesnt need to be its own function, just put it on line 20

public/js/calendarPage.js
```diff
@@ -0,0 +1,45 @@
...
+
+    async function addCalendarEvents() {
+        const response = await fetch("./api/calendar");
+        const events = await extractResponseData(response);
```
mporam, line 20: this could just be `const events = await response.json()`

src/Controllers/API/GetAllCalendarEventsController.php
```diff
@@ -0,0 +1,34 @@
...
+        $this->eventModel = $eventModel;
+    }
+
+    /**
+     * Checks for logged-in status,
+     * gets all events from DB
+     */
+    public function __invoke(Request $request, Response $response, array $args): Response
+    {
+        $statusCode = 200;
```
mporam, line 25: why? just put the `200` in the json response. dont need a variable

templates/partials/layout/nav.phtml
```diff
@@ -26,6 +26,7 @@
         <ul class="dropdown-menu">
             <li><a href="/addEvent">New Event</a></li>
             <li><a href="/events-past">Past Events</a></li>
+            <li><a href="/calendar">Calendar</a></li>
```
mporam, line 29: can we have a separater between this link to differentiate it from the event pages (see the applicants drop down). maybe also put it first in the list?

mporam (changes requested):

public/js/calendarPage.js
```diff
@@ -1,45 +1,35 @@
...
+  events.map((event) => {
+    if (event.categoryId == 1) {
+      event.color = "#4eab1a";
+    } else if (event.categoryId == 2) {
+      event.color = "#002aff";
+    } else if (event.categoryId == 3) {
+      event.color = "#b5b601";
+    } else if (event.categoryId == 4) {
+      event.color = "#c03d3d";
+      event.allDay = true;
```
mporam, line 23: this feels like the kind of thing that should be done on the back end. or with some logic that says if there is no start time its an all day event. the event IDs you have currently wont necessarily match the ones on the live application, so this likely wont work

mporam (changes requested):
still other outstanding comments too

templates/partials/layout/nav.phtml
ndrj19: isn't that there already?

mporam (changes requested):

mporam (approved):

PR #568, opened 2023-11-23
S7 add course types brainworms

jacobl992:


public/js/addCourseCategory.js
```diff
@@ -0,0 +1,104 @@
+const categoryForm = document.querySelector('form');
+const message = document.querySelector('#messages');
+//const categoryInput = document.querySelector('#addCategory');
```
mporam, line 3: if not needed, please remove

public/js/addCourseCategory.js
```diff
@@ -0,0 +1,104 @@
...
+                }
+            })
+    }
+});
+
+/**
+ * Adds data from form into an object with the field name as key and the form value as value.
+ */
+let getCompletedFormData = () => {
+    let data = {
```
mporam, line 67: why not just return the object, you dont need the data variable

public/js/addCourseCategory.js
```diff
@@ -0,0 +1,104 @@
...
+        courseCategory: categoryForm.elements['addCategory'].value
+    }
+    return data;
+}
+
+let validateCategoryInput = (data) => {
+
+    validate = {
```
mporam, line 75: same as above, dont need the validate variable, just return the object

public/js/addCourseCategory.js
```diff
@@ -0,0 +1,104 @@
...
+            break;
+        case 'isName':
+            htmlString = `Please use alphanumeric characters only.`;
+            break;
+        default:
+            htmlString = `This field is invalid.`;
+            break;
+    }
+    return htmlString;
+}
```
mporam, line 104: does this function not already exist somewhere else? can you just use that one?

public/js/deleteCategory.js
```diff
@@ -0,0 +1,21 @@
...
+                body: JSON.stringify({"id": id}),
+                headers: {
+                    "Content-Type": "application/json"
+                }
+            });
+            const responseData = await response.json();
+            document.cookie = `response = ${responseData.message}`;
+            window.location.reload();
+        } catch (error) {
+            console.error('Error', error);
```
mporam, line 18: surely we need to show the user an error, rather than just putting something in the console?

public/js/deleteCategory.js
```diff
@@ -0,0 +1,21 @@
...
+        try {
+            const response = await fetch('/api/deleteCategory', {
+                method: 'POST',
+                body: JSON.stringify({"id": id}),
+                headers: {
+                    "Content-Type": "application/json"
+                }
+            });
+            const responseData = await response.json();
+            document.cookie = `response = ${responseData.message}`;
```
mporam, line 15: why? cant see you using this message anywhere?

public/js/deleteCategory.js
```diff
@@ -0,0 +1,21 @@
...
+            window.location.reload();
```
mporam, line 16: not a fan of this but it is consistent with the rest of the app (unfortunately). however you need to check the response is successful before reloading

src/Controllers/API/AddCourseCategoryController.php
```diff
@@ -0,0 +1,51 @@
...
+        $responseData = [
+            'success' => false,
+            'message' => 'Unexpected Error.',
+            'data' => []
+        ];
+        $statusCode = 400;
+
+        try {
+                $newCategory = CourseCategorySanitiser::sanitise($newCategory);
+                $insertedId = $this->courseModel->addCategory($newCategory);
```
mporam, line 36: incorrect indentation

src/Controllers/API/AddCourseCategoryController.php
```diff
@@ -0,0 +1,51 @@
...
+        } catch (\Exception $exception) {
+            $responseData['message'] = $exception->getMessage();
```
mporam, line 38: if the query fails because the DB crashed, this will send a 400 status code but with a DB broken error, which makes no sense. you need a catch for PDO exception, to set the status to 500, and a catch for the sanitise exceptions to keep the 400 status code

src/Controllers/API/DeleteCourseCategoryController.php
```diff
@@ -0,0 +1,45 @@
...
+        $deletedCategory = $data['id'];
+
+        $result = $this->courseModel->deleteCategory($deletedCategory);
+
+        if ($result) {
+            $data = [
+                    'success' => true,
+                    'message' => 'Category deleted',
+                    'id' => [$deletedCategory]
+                     ];
```
mporam, line 35: incorrect indentation

src/Controllers/API/DeleteCourseCategoryController.php
```diff
@@ -0,0 +1,45 @@
...
+            return $this->respondWithJson($response, $data);
+        } else {
+            $data = [
+                    'success' => false,
+                    'message' => 'Error - please contact administrator'
```
mporam, line 40: probably `Unexpected error occurred` would be better here

src/Controllers/FrontEnd/AddNewCategoryPageController.php
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Checks for logged-in status,
+     * gets courses categories from DB
+     * and returns rendered landing screen for Courses page
+     */
+    public function __invoke(Request $request, Response $response, array $args): Response
+    {
+        if (!empty($_SESSION['loggedIn']) && $_SESSION['loggedIn']) {
+            //will change this
```
mporam, line 30: either change it, or remove this comment

templates/addCourse.phtml
```diff
@@ -26,6 +26,12 @@ $trainersCheckbox = \Portal\ViewHelpers\TrainerCheckboxViewHelper::displayTraine
         <label for="course-name">Course Name</label>
         <input name="courseName" class="create-courses form-control" type="text" placeholder="Course name" data-required="true" data-max='255' maxlength="255" />
         <div id="courseNameError" class="alert hidden formItem_alert"></div>
+        <label for="course-category">Category</label>
```
mporam, line 29: for attribute has to match the ID of the select, remove the `-` and capitalise `Category`

mporam (changes requested):

public/js/deleteCategory.js
```diff
@@ -12,10 +13,13 @@ deleteButtons.forEach(deleteCatBtn => {
                 }
             });
             const responseData = await response.json();
-            document.cookie = `response = ${responseData.message}`;
-            window.location.reload();
+            if (responseData.success) {
+                window.location.reload();
+            } else {
+                deleteError.classList.remove('hidden');
```
mporam, line 19: the response should have an error message in it that says what went wrong, perhaps display that error here rather than a generic one?

mporam (changes requested):

mporam (approved):

mporam (approved):

PR #574, opened 2024-04-22
radio buttons added and courses table in correct format

maxeh987:
https://trello.com/c/RxGs7Wsc/197-story-4-change-in-person-remote-to-radio-buttons

added name to the radio buttons so they are paired
added js so that both radio buttons affect the 'hiddenness' of the spaces boxes
changed the logic in create courses table loop function to get $location rather than $remote or $in_person.


public/js/editApplicant.js
```diff
@@ -45,3 +45,6 @@ function aptitudeScoreButtonClick(e)
                 showAlertforApptitudeButton('#aptitudeScoreError')
         });
 }
+
```
elizabeth-hartley, line 48: Delete these 3 empty lines :)

elizabeth-hartley (changes requested):
Just those 3 empty lines

public/js/addCourse.js
```diff
@@ -6,8 +6,16 @@ const in_person_spaces = document.querySelector('#in_person_spaces');
 const remote_spaces = document.querySelector('#remote_spaces');
 const courseCategory = document.querySelector('#courseCategory');
 
-in_person_checkbox.addEventListener('change', () => in_person_spaces.classList.toggle('hidden'));
-remote_checkbox.addEventListener('change', () => remote_spaces.classList.toggle('hidden'));
+// if (in_person_spaces.classList.contains('hidden') || remote_spaces.classList.contains('hidden')) {
+//
+// }
```
matt-ioa, line 11: Get rid of these comments

templates/addCourse.phtml
```diff
@@ -48,14 +48,12 @@ $trainersCheckbox = \Portal\ViewHelpers\TrainerCheckboxViewHelper::displayTraine
 
         <div class="form-group">
             <label for="in_person">In Person?</label>
-            <input type="checkbox" id="in_person" value="1" class="create-courses" name="in_person">
-            <input id="in_person_spaces" name="in_person_spaces" class="create-courses form-control hidden" type="number" placeholder="How many spaces available?" min="1" max="999" data-required="true" />
+            <input type="radio" id="in_person" value="1" class="create-courses" name="radio">
```
matt-ioa, line 51: "radio" is not a good name for these two inputs, because that will become the key for the data that's submitted. Something that describes what you're selecting would be better.

public/js/addCourse.js
```diff
@@ -6,8 +6,16 @@ const in_person_spaces = document.querySelector('#in_person_spaces');
...
+
+in_person_checkbox.addEventListener('click', () => {
```
matt-ioa, line 13: Maybe now you should rename these two consts `in_person_radio` and `remote_radio` now that they're both radio buttons.

matt-ioa (changes requested):
Please add code to make sure the new radio buttons actually work

JBHarrison5:
tested and it works

matt-ioa (approved):

PR #578, opened 2024-04-24
S8t1

maxeh987:


public/js/addOrEditCourse.js
matt-ioa, line 1: I really don't like this approach of having adding/editing in one js file. I think it would be better to have two separate files, even if it means duplicating code. It would just be so much clearer than having that weird 0/1 situation.

src/Controllers/API/EditCourseController.php
```diff
@@ -0,0 +1,24 @@
...
+
+    public function __construct(CourseModel $courseModel)
+    {
+        $this->courseModel = $courseModel;
+    }
+
+    public function __invoke(RequestInterface $request, ResponseInterface $response)
+    {
+        $datum = $request->getParsedBody();
+        $this->courseModel->updateCourse($datum);
```
matt-ioa, line 21: Please don't call this `$datum`, that's an awful name. `$courseData` would be better.

src/Controllers/FrontEnd/EditCoursePageController.php
```diff
@@ -0,0 +1,42 @@
...
+        $this->renderer = $renderer;
+        $this->trainerModel = $trainerModel;
+    }
+
+    public function __invoke(Request $request, Response $response, array $args): Response
+    {
+
+        if (!empty($_SESSION['loggedIn']) && $_SESSION['loggedIn']) {
+            $id = $args['id'];
+            if ($id > 0) {
```
matt-ioa, line 30: `$id` is actually going to be a string at this point, so this comparison doesn't really make sense

src/ViewHelpers/AddCourseViewHelper.php
```diff
@@ -4,12 +4,15 @@
 
 class AddCourseViewHelper
 {
-    public static function courseCategoryDropdown(array $categories): string
+    public static function courseCategoryDropdown(array $categories, $courseCatId): string
```
matt-ioa, line 7: Is `$courseCatId` an int or string?

src/ViewHelpers/AddCourseViewHelper.php
```diff
@@ -4,12 +4,15 @@
...
     {
         $courseCategoryDropdown = '';
         foreach ($categories as $category) {
+            $selected = $category['id'] === $courseCatId ? 'selected' : '';
             $courseCategoryDropdown .=
-                '<option value="' . $category['id'] . '">' . $category['category'] . '</option>';
+                '<option selected="' . $selected . '" value="' . $category['id'] . '">'
+                . $category['category'] . '</option>';
+            var_dump($category);
```
matt-ioa, line 15: You've still got a `var_dump` here

src/ViewHelpers/TrainerCheckboxViewHelper.php
```diff
@@ -9,17 +9,29 @@ class TrainerCheckboxViewHelper
     /**
      * take list of trainers and foreach to display them as individual checkbox items
      */
-    public static function displayTrainerCheckbox(array $trainers): string
+    public static function displayTrainerCheckbox(array $trainers, $courseTrainers = []): string
```
matt-ioa, line 12: Add an array type hint here

templates/addCourse.phtml
```diff
@@ -58,6 +57,7 @@ $trainersCheckbox = \Portal\ViewHelpers\TrainerCheckboxViewHelper::displayTraine
         </div>
 
         <div id="formGroupError" class="alert hidden formItem_alert"></div>
+        <input type="hidden" value="1" name="editOrAdd">
```
matt-ioa, line 60: I still think this is a terrible idea!

templates/courseCategories.phtml
```diff
@@ -26,7 +26,7 @@ require 'partials/layout/nav.phtml';
             <th class="col-xs-9">Category</th>
             <th class="col-xs-3">Action</th>
         </tr>
-        <?php echo Portal\ViewHelpers\CourseCategoriesViewHelper::displayCourseCategories($categories); ?>
+        <?= Portal\ViewHelpers\CourseCategoriesViewHelper::displayCourseCategories($categories); ?>
```
matt-ioa, line 29: It would have been way better to find/replace all these tags in a separate PR, so it would be easier to find the lines that are relevant only to this one!

matt-ioa (changes requested):
Some changes requested please

PR #583, opened 2024-04-26
S7 t2delete course modal

elizabeth-hartley:
All peer review comments rectified and composer sniff and composer test done


public/js/deleteCourseModal.js
```diff
@@ -0,0 +1,59 @@
+let deleteButtons = document.querySelectorAll('.btnDeleteCourse');
+for (let db of deleteButtons) {
+    db.addEventListener("click", () => {
+        const courseId = db.getAttribute('data-courseId');
+        displayDetails('./api/deleteCourse/' + courseId, courseId);
+    });
+}
+
+function displayDetails(url, courseId) {
```
matt-ioa, line 9: It would be better to define this function before it's used, not after

public/js/deleteCourseModal.js
```diff
@@ -0,0 +1,59 @@
...
+    fetch(url).then(
+        response => {
+            return response.json();
+        }).then(data => {
+            let studentNamesContainer = document.querySelector('#studentNamesContainer');
+            studentNamesContainer.innerHTML = '<h4>Name</h4>';
+            let DeleteCoursesForm = document.querySelector('#DeleteCoursesForm');
+            DeleteCoursesForm.innerHTML = '';
+            DeleteCoursesForm.action = "api/deleteCourse/" + courseId;
```
matt-ioa, line 18: Isn't this pretty much the same as the `url`? So maybe `courseId` isn't needed at all?

public/js/deleteCourseModal.js
```diff
@@ -0,0 +1,59 @@
...
+            let prettyCourses = makeCoursesPretty(data['availableCourses'])
+            for (let student of data['applicants']) {
+                generateStudents(student['name'], studentNamesContainer);
+                generateOptions(prettyCourses, DeleteCoursesForm, student['id'], data['availableCourses']);
```
matt-ioa, line 22: Again, better to define both `generateStudents` and `generateOptions` before they're called here. It makes the code more readable and saves people searching for where the functions are defined.

public/js/deleteCourseModal.js
```diff
@@ -0,0 +1,59 @@
...
+    )
+}
+
+function generateStudents(name, studentNamesContainer) {
+    let studentName = document.createElement('h5');
+    studentName.innerText = name;
+    studentNamesContainer.appendChild(studentName);
+}
+
+function makeCoursesPretty(courses) {
```
matt-ioa, line 34: This name doesn't really describe what the function does. Maybe something like `formatCourseNameAndStart` is less exciting but more descriptive!

src/Controllers/API/DeleteCourseController.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        $this->courseModel = $courseModel;
+    }
+
+    public function __invoke(Request $request, Response $response, array $args): Response
+    {
+        $students = $request->getParsedBody();
+        $courseToDelete = $args['id'];
+        foreach ($students as $studentId => $value) {
+            if ($value === '0') {
```
matt-ioa, line 25: Could you add a comment to explain what this 0 value represents?

matt-ioa (changes requested):
Just a couple of changes requested

src/Controllers/API/DeleteCourseController.php
elizabeth-hartley: If no course is assigned to an applicant, it gives the applicant the value of 0 and they are removed from all courses

public/js/deleteCourseModal.js
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+    )
+}
+
+let deleteButtons = document.querySelectorAll('.btnDeleteCourse');
+for (let db of deleteButtons) {
+    db.addEventListener("click", () => {
+        const courseId = db.getAttribute('data-courseId');
+        displayDetails('./api/deleteCourse/' + courseId, courseId);
```
matt-ioa, line 48: How does this still work when `displayDetails` only takes a url now?

matt-ioa (approved):

PR #594, opened 2024-09-24
Edit trainer page t3

joseph-padfield:
Shows edit trainer form page, populated with trainer details


templates/editTrainer.phtml
tking42, line 1: in the forms change placeholder to value = so that can edit the trainers details directly

